PR #12, opened 2022-02-17
Add JSONLD examples and SHACL schema for C01 credit classes, credit batches and projects

wgwz:
## Description

Closes: https://github.com/regen-network/regen-registry/issues/780

The most critical file to review is this one: `shacl/credit-batches/credit-batch.ttl`. I decided to create new shape definitions just for the sake of getting something that works. But I believe we might want to dedupe and move some of these new shape definitions to pre-existing shape definitions. I.e. I define a new `ProjectShape` in `credit-batch.ttl` even though there is already a `ProjectShape` defined in `project-shape.ttl`.

I confirmed that the example JSON-LD conforms with the new credit batch spec:

```
kyle@lookfar4:~/regen/regen-registry-standards$ pyshacl -s shacl/credit-batches/credit-batch.ttl jsonld/credit-batches/C01-verified-carbon-standard.json
Validation Report
Conforms: True
```

I think that I might move `shacl/credit-batches/credit-batch.ttl` into `shacl/credit-batches/C01-verified-carbon-standard.ttl`. I'm not really sure what the best name for this file is, I could use some feedback there as well.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


wgwz:
I just pushed a new commit that added an example JSON-LD based on the [Mai Nadombe](https://regen-registry.gitbook.io/credit-classes/group-1/credit-class-metadata/mai-nadombe-credit-batch-example) example. This newly add project graph data also validates with the `shacl/credit-batches/credit-batch.ttl` schema (since I defined a `Project` schema there)

Showing the validation result:
```
kyle@lookfar4:~/regen/regen-registry-standards$ pyshacl -s shacl/credit-batches/credit-batch.ttl jsonld/projects/C01-verified-carbon-standard.json
Validation Report
Conforms: True
```

blushi:
> I think that I might move shacl/credit-batches/credit-batch.ttl into shacl/credit-batches/C01-verified-carbon-standard.ttl. I'm not really sure what the best name for this file is, I could use some feedback there as well.

I think that would make sense since here we're defining constraints for credit batches that will belong to this particular credit class (C01-verified-carbon-standard)

shacl/credit-batches/credit-batch.ttl
```diff
@@ -0,0 +1,123 @@
...
+    sh:path schema:name ;
+    sh:datatype xsd:string ;
+  ] ;
+  sh:property [
+    sh:path schema:url ;
+    sh:datatype schema:URL ;
+  ] ;
+.
+
+regen:VcsMethodologyShape a sh:NodeShape ;
```
blushi, line 101: we could just re-use `regen:ProjectActivityShape` here and rename it to a more general name

wgwz: I think I did this correctly!

jsonld/credit-batches/C01-verified-carbon-standard.json
```diff
@@ -7,10 +7,10 @@
     "xsd": "http://www.w3.org/2001/XMLSchema#"
   },
   "@type": "regen:CreditBatch",
-  "regen:vcsRetirementSerialNumber": "",
+  "regen:vcsRetirementSerialNumber": "abcdefg",
```
blushi, line 10: could you make sure to revert the changes in this file (it's supposed to be a template with empty strings and 0 values)

shacl/credit-batches/C01-verified-carbon-standard.ttl
```diff
@@ -0,0 +1,147 @@
...
+  ] ;
+  sh:property [
+    sh:path regen:batchVerificationReport ;
+    sh:datatype schema:URL ;
+    sh:minCount 1 ;
+    sh:maxCount 1 ;
+  ] ;
+  sh:property [
+    sh:path regen:project ;
+    sh:node regen:Project ;
```
blushi, line 38: you have to specify the shape name here, not the class name
```suggestion
    sh:node regen:ProjectShape ;
```

shacl/credit-batches/C01-verified-carbon-standard.ttl
```diff
@@ -0,0 +1,147 @@
...
+  ] ;
+  sh:property [
+    sh:path regen:vcsProjectPage ;
+    sh:datatype schema:URL ;
+    sh:minCount 1 ;
+    sh:maxCount 1 ;
+  ] ;
+  sh:property [
+    sh:path regen:projectProponent ;
+    sh:node regen:Organization ;
```
blushi, line 60: ditto
```suggestion
    sh:node regen:OrganizationShape ;
```

shacl/credit-batches/C01-verified-carbon-standard.ttl
```diff
@@ -0,0 +1,147 @@
...
+  sh:property [
+    sh:path regen:vcsProjectType ;
+    sh:minCount 1 ;
+    sh:maxCount 1 ;
+    sh:minLength 1 ;
+    sh:datatype xsd:string ;
+  ] ;
+  sh:property [
+    sh:path regen:projectActivity ;
+    sh:node regen:NameUrl ;
```
blushi, line 73:
```suggestion
    sh:node regen:NameUrlShape ;
```

shacl/credit-batches/C01-verified-carbon-standard.ttl
```diff
@@ -0,0 +1,147 @@
...
+  ] ;
+  sh:property [
+    sh:path regen:vcsMethodology ;
+    sh:node regen:NameUrl ;
```
blushi, line 77:
```suggestion
    sh:node regen:NameUrlShape ;
```

shacl/credit-batches/C01-verified-carbon-standard.ttl
```diff
@@ -0,0 +1,147 @@
...
+  ] ;
+  sh:property [
+    sh:path regen:size ;
+    sh:node regen:Size ;
```
blushi, line 81:
```suggestion
    sh:node regen:SizeShape ;
```

blushi (changes requested):
It seems like you're not using the same url for `qudt` namespace in the SHACL and the JSON-LD data, which gives you some unexpected behavior when running validation.
Let's just start using `http://qudt.org/schema/qudt` everywhere which now seems the most up-to-date one and we don't need to worry about the version.

> The [schema](https://github.com/regen-network/regen-registry-standards/blob/kyle/780-offcain-metadata/shacl/credit-batches/C01-verified-carbon-standard.ttl#L140-L145) for this piece of the graph is this:
> 
> ```
>   sh:property [
>     sh:path qudt:numericValue ;
>     # TODO: This is not validating as expected (I think..)
>     # See "qudt:numericValue" in
>     # jsonld/credit-batches/C01-verified-carbon-standard.json
>     sh:datatype xsd:double ;
> ```
> 
> I'm expecting an error because the test data is a string value, and not a double. Does the `xsd:double` datatype just not have any constraints? Do we need to add constraints?

SHACL only verifies that the `@type` is `xsd:double`, not the actual value. You could also represent a double using a string, e.g. "380.16". In addition, you could also add `sh:minExclusive 0 ;` to double check it's a positive number.

wgwz:
> It seems like you're not using the same url for `qudt` namespace in the SHACL and the JSON-LD data, which gives you some unexpected behavior when running validation. Let's just start using `http://qudt.org/schema/qudt` everywhere which now seems the most up-to-date one and we don't need to worry about the version.

Ah, ok. I see. I've updated that and made that change in all the places it was used (it was only two instances).

> SHACL only verifies that the `@type` is `xsd:double`, not the actual value. You could also represent a double using a string, e.g. "380.16". In addition, you could also add `sh:minExclusive 0 ;` to double check it's a positive number.

I see, ok, this is what I was getting confused by. The interesting thing is that it seems like some types do some light validation to the data when verifying with a SHACL schema. For example, the date type, when I ran the SHACL schema, it raised an issue if it wasn't actually a date.

I also removed all the fake data I had in sample JSON-LD's, it's all empty strings or 0's now.

blushi:
> I see, ok, this is what I was getting confused by. The interesting thing is that it seems like some types do some light validation to the data when verifying with a SHACL schema. For example, the date type, when I ran the SHACL schema, it raised an issue if it wasn't actually a date.

I think that if you just have a string like "foo" for some xsd:double, the validation should fail still.
I believe your issue was just that the validation wasn't run because you were using the type name as `sh:node` instead of the shape name.

wgwz:
> After some afterthought and some testing on that, if you just have a string like "foo" for some xsd:double, the validation should actually fail.
> 
> I believe your issue was rather that the validation wasn't run because you were using the type name as `sh:node` instead of the shape name.

So I took a look at this, and I tried testing the change to "foo" for xsd:double value. With just that change, I was still not getting any validation error. However with the patch below, I was able to get it to raise a validation error. Should I commit these and push them?

TL;DR I replaced `qudt:numericValue` with `xsd:double` in both the SizeShape definition and in the example JSON-LD.

```
diff --git a/jsonld/credit-batches/C01-verified-carbon-standard.json b/jsonld/credit-batches/C01-verified-carbon-standard.json
index 61f5499..2fde0c6 100644
--- a/jsonld/credit-batches/C01-verified-carbon-standard.json
+++ b/jsonld/credit-batches/C01-verified-carbon-standard.json
@@ -57,7 +57,7 @@
         "@type": "qudt:Unit",
         "@value": "unit:HA"
       },
-      "qudt:numericValue": {
+      "xsd:double": {
         "@type": "xsd:double",
         "@value": ""
       }
diff --git a/shacl/credit-batches/C01-verified-carbon-standard.ttl b/shacl/credit-batches/C01-verified-carbon-standard.ttl
index d4ec67f..fb6ee5a 100644
--- a/shacl/credit-batches/C01-verified-carbon-standard.ttl
+++ b/shacl/credit-batches/C01-verified-carbon-standard.ttl
@@ -138,7 +138,7 @@ regen:SizeShape a sh:NodeShape ;
     sh:datatype qudt:Unit ;
   ] ;
   sh:property [
-    sh:path qudt:numericValue ;
+    sh:path xsd:double ;
     sh:minExclusive 0 ;
     sh:datatype xsd:double ;
   ] ;
```

Here's the validation error for when it's a negative value:
```
Constraint Violation in MinExclusiveConstraintComponent (http://www.w3.org/ns/shacl#MinExclusiveConstraintComponent):
        Severity: sh:Violation
        Source Shape: [ sh:datatype xsd:double ; sh:minExclusive Literal("0", datatype=xsd:integer) ; sh:path xsd:double ]
        Focus Node: [ <qudt:unit> Literal("unit:HA" = None, datatype=<qudt:Unit>) ; rdf:type regen:Size ; xsd:double Literal("-12.0", datatype=xsd:double)
 ]
        Value Node: Literal("-12.0", datatype=xsd:double)
        Result Path: xsd:double
        Message: Value is not > Literal("0", datatype=xsd:integer)
```

And here's the validation error for "foo":
```
Constraint Violation in MinExclusiveConstraintComponent (http://www.w3.org/ns/shacl#MinExclusiveConstraintComponent):
        Severity: sh:Violation
        Source Shape: [ sh:datatype xsd:double ; sh:minExclusive Literal("0", datatype=xsd:integer) ; sh:path xsd:double ]
        Focus Node: [ <qudt:unit> Literal("unit:HA" = None, datatype=<qudt:Unit>) ; rdf:type regen:Size ; xsd:double Literal("foo" = None, datatype=xsd:double) ]
        Value Node: Literal("foo" = None, datatype=xsd:double)
        Result Path: xsd:double
        Message: Value is not > Literal("0", datatype=xsd:integer)
```

blushi:
> TL;DR I replaced `qudt:numericValue` with `xsd:double` in both the SizeShape definition and in the example JSON-LD.

The path value shouldn't have any impact on the validation. I have tried using just `qudt:numericValue` as path and I get the same validation errors.

jsonld/credit-classes/C01-verified-carbon-standard-class.json
```diff
@@ -0,0 +1,50 @@
...
+    "schema": "http://schema.org/",
+    "regen": "http://regen.network/",
+    "qudt": "http://qudt.org/schema/qudt",
+    "unit": "http://qudt.org/2.1/vocab/unit",
+    "xsd": "http://www.w3.org/2001/XMLSchema#"
+  },
+  "@type": "regen:CreditClass",
+  "regen:description": [
+      "This credit class provides a vehicle for nature based Verified Carbon Units (VCUs) to enter the blockchain space via issuance on Regen Ledger. It can be used by project developers, credit brokers, and other stakeholders interested in digitizing carbon credits issued by the VCS program to make them available in the emerging world of decentralized finance."
+  ],
```
S4mmyb, line 12:
```suggestion
  "regen:description": "",
```

jsonld/credit-classes/C01-verified-carbon-standard-class.json
```diff
@@ -0,0 +1,50 @@
...
+  "regen:sourceRegistry": {
+    "@type": "regen:Organization",
+    "schema:name": "Verra"
```
S4mmyb, line 15:
```suggestion
    "schema:name": ""
```

jsonld/credit-classes/C01-verified-carbon-standard-class.json
```diff
@@ -0,0 +1,50 @@
...
+  },
+  "regen:sectoralScope": {
+    "@type": "schema:ItemList",
+    "itemListElement": [
+      "Agriculture Forestry and Other Land Use (AFOLU)",
+      "Livestock and Manure Management"
+    ]
```
S4mmyb, line 22:
```suggestion
    "itemListElement": []
```

jsonld/credit-classes/C01-verified-carbon-standard-class.json
```diff
@@ -0,0 +1,50 @@
...
+  },
+  "regen:offsetGenerationMethod": {
+    "@type": "schema:ItemList",
+    "itemListElement": [
+      "Reduction",
+      "Removal"
+    ]
```
S4mmyb, line 29:
```suggestion
    "itemListElement": []
```

jsonld/credit-classes/C01-verified-carbon-standard-class.json
```diff
@@ -0,0 +1,50 @@
...
+    {
+      "@type": "regen:vcsMethodology",
+      "schema:name": "name",
+      "schema:URL": {
+        "@type": "schema:URL",
+        "value": "google.com"
+      }
+    }
+  ],
+  "regen:verificationMethod": "third-party verified"
```
S4mmyb, line 49:
```suggestion
  "regen:verificationMethod": ""
```

S4mmyb (commented):
I mostly reviewed for the JSON-LD content and made a few minor suggestions to omit value content (ie description, organization, scope, and offset method values) in the JSONLD files and instead leave all fields as null/empty. 

After some discussion with @clevinson I think we want to use the `regen-registry-standards` repo to hold templates and standards for graphs and validations, but not the actual content for credit classes, batches, projects, etc... Instead we will start a new new repository, `regen-registry-ops`, where we can keep all the implemented credit classes, projects, and credit batches json-ld files, as well as many of the scripts in our multisig ops, credit class issuances, etc...

jsonld/projects/C01-verified-carbon-standard-projects.json
```diff
@@ -0,0 +1,50 @@
...
+  "regen:vcsProjectPage": {
+    "@type": "schema:URL",
+    "@value": ""
+  },
+  "regen:projectProponent": {
+    "@type": "regen:Organization",
+    "schema:name": ""
+  },
+  "regen:vcsProjectType": "",
+  "regen:projectActivity": "",
```
blushi, line 23: we should provide `schema:name` and `schema:url` here
```suggestion
  "regen:projectActivity": {
   "schema:name": "",
    "schema:url": {
      "@type": "schema:URL",
      "@value": ""
    }
  }
```

shacl/credit-batches/C01-verified-carbon-standard.ttl
```diff
@@ -0,0 +1,145 @@
...
+    sh:datatype schema:URL ;
+    sh:minCount 1 ;
+    sh:maxCount 1 ;
+  ] ;
+  sh:property [
+    sh:path regen:project ;
+    sh:node regen:ProjectShape ;
+    sh:minCount 1 ;
+    sh:maxCount 1 ;
+  ] ;
```
blushi, line 41: we should move the project shape to its own file now that it's not part of the credit batch metadata anymore

jsonld/credit-batches/C01-verified-carbon-standard.json
```diff
@@ -0,0 +1,19 @@
+{  "@context": {
```
blushi, line 1: let's rename this file to "C01-verified-carbon-standard-batch.json"?

blushi (changes requested):
it'd be nice to update the credit batch shacl schema and create new ones based on the project and credit class metadata jsonld templates

shacl/credit-classes/credit-class.ttl
```diff
@@ -16,6 +16,12 @@ regen:CreditClassVersionShape
     sh:path schema:url ;
     sh:datatype schema:URL ;
   ] ;
+  sh:property [
+    sh:path regen:standard ;
+    sh:maxCount 1 ;
+    sh:minCount 1 ;
+    sh:node regen:StandardShape ;
+  ] ;
```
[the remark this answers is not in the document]
wgwz: Ah, ok, I’ll fix that. I was confused during the merge process with the main branch and I got the direction of that wrong.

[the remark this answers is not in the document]
S4mmyb: thanks @wgwz!

jsonld/credit-classes/C01-verified-carbon-standard-class.json
```diff
@@ -0,0 +1,50 @@
...
+  "regen:offsetGenerationMethod": {
+    "@type": "schema:ItemList",
+    "itemListElement": [
+      "Reduction",
+      "Removal"
+    ]
+  },
+  "regen:approvedMethodologies": [
+    {
+      "@type": "regen:vcsMethodology",
```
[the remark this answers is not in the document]
S4mmyb: good point, thanks @blushi!

jsonld/credit-classes/C01-verified-carbon-standard-class.json
```diff
@@ -0,0 +1,50 @@
...
+      "schema:name": "name",
+      "schema:URL": {
+        "@type": "schema:URL",
+        "value": "google.com"
```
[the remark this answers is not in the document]
S4mmyb: @blushi is there a reason we're doing reason to do 
`"schema:URL": {
        "@type": "schema:URL",
        "value": "
`

over just 

`"schema:URL": ""`?

Shouldn't it just be something more along the lines of the following? Or am I missing something?
 
`"methodologyLink": {
        "@type": "schema:URL",
        "value": ""
`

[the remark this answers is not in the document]
blushi: the field name is `schema:url` (not `schema:URL` which is the type)
see difference between the 2: https://schema.org/url (property) vs https://schema.org/URL (datatype)

jsonld/credit-batches/C01-verified-carbon-standard.json
```diff
@@ -0,0 +1,19 @@
...
+    "@value": ""
+  },
+  "regen:batchMonitoringReport": {
+    "@type": "schema:URL",
+    "@value": ""
+  },
+  "regen:batchVerificationReport": {
+    "@type": "schema:URL",
+    "@value": ""
+  }
```
blushi, line 18: we had also discussed adding the on chain project id info to the credit batch metadata for easier reference as long as on-chain projects are not live @S4mmyb

shacl/credit-batches/C01-verified-carbon-standard.ttl
wgwz: I think I moved this into a decent location, but we always adjust later if needed. I moved it `shacl/projects/C01-verified-carbon-standard-project.ttl` for now.

shacl/credit-classes/credit-class.ttl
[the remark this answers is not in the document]
wgwz: @blushi I believe I fixed this issue. I made this merge process much more more complicated than it needed to be. :sweat_smile:

shacl/credit-batches/C01-verified-carbon-standard.ttl
wgwz: Going to mark this as resolved for now.

wgwz:
> I mostly reviewed for the JSON-LD content and made a few minor suggestions to omit value content (ie description, organization, scope, and offset method values) in the JSONLD files and instead leave all fields as null/empty.
> 
> After some discussion with @clevinson I think we want to use the `regen-registry-standards` repo to hold templates and standards for graphs and validations, but not the actual content for credit classes, batches, projects, etc... Instead we will start a new new repository, `regen-registry-ops`, where we can keep all the implemented credit classes, projects, and credit batches json-ld files, as well as many of the scripts in our multisig ops, credit class issuances, etc...

@S4mmyb @blushi  One point I wanted raise here is that it would be helpful to have testing data (JSONLD with non-empty values) for this repo. I think that when you start using SHACL schema for validation, it's important to have a simple way to test that the validations are working correctly. I have some ideas for how we could potentially set up a CI for this repo to run tests. I'm uncertain of what will be in `regen-registry-ops` and how or if we could use it for these purposes.

shacl/credit-classes/credit-class.ttl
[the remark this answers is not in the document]
wgwz: Going to resolve for now

wgwz:
> it'd be nice to update the credit batch shacl schema and create new ones based on the project and credit class metadata jsonld templates

@blushi I've setup two SHACL schema for project and credit class (C01):

- `C01-verified-carbon-standard-project.ttl`
- `C01-verified-carbon-standard-class.ttl`

The project schema was pulled out of the batch schema we set up previously, and I only made a small adjustment.

I think the most import to review at this point are `C01-verified-carbon-standard-class.ttl` and the corresponing `C01-verified-carbon-standard-class.json`. I'm not 100% sure how we should structure the JSON-LD, specifically the things that lists, and depending on that, maybe we need some updates to it's SHACL.

jsonld/credit-batches/C01-verified-carbon-standard.json
blushi: up @wgwz @S4mmyb

jsonld/credit-classes/C01-verified-carbon-standard-class.json
```diff
@@ -0,0 +1,43 @@
...
+    "xsd": "http://www.w3.org/2001/XMLSchema#"
+  },
+  "@type": "regen:CreditClass",
+  "schema:description": "",
+  "regen:sourceRegistry": {
+    "@type": "regen:Organization",
+    "schema:name": ""
+  },
+  "regen:sectoralScope": {
+    "@type": "schema:ItemList",
```
blushi, line 14: I'm not sure whether we should be using `schema:ItemList` or not. I think it would only makes sense if we want to plan some other ItemList related properties along with it, like https://schema.org/itemListOrder or https://schema.org/numberOfItems
Previously, for representing lists, we've just been using this: https://github.com/regen-network/regen-registry-standards/blob/cdca0cfef18657d7b5b2b309429aeaf2caa603ac/shacl/project-page.ttl#L227
and then an example JSON-LD would look like (https://json-ld.org/spec/latest/json-ld/#sets-and-lists): 
```
"http://regen.network/galleryPhotos": {
  "@list": [
    {
      "@type": "http://schema.org/URL",
      "@value": "https://regen-registry.s3.amazonaws.com/projects/woodburn/image1.jpg"
    },
    {
      "@type": "http://schema.org/URL",
      "@value": "https://regen-registry.s3.amazonaws.com/projects/woodburn/image2.png"
    },
    {
      "@type": "http://schema.org/URL",
      "@value": "https://regen-registry.s3.amazonaws.com/projects/woodburn/image3.png"
    },
    {
      "@type": "http://schema.org/URL",
      "@value": "https://regen-registry.s3.amazonaws.com/projects/woodburn/image4.png"
    }
  ]
}
```

cc/ @wgwz @S4mmyb

wgwz: @S4mmyb could you provide some example data for what will be in the `sectoralScope` list for the credit class json-ld? Will it be a list of strings? I'm not familiar with this terminology, would also be curious to learn more about that :)

I think unless there's a good reason to, we should probably stick to what's been done already, in terms of using `dash:ListShape`

jsonld/credit-batches/C01-verified-carbon-standard.json
wgwz: I would be glad to add that here. Did you mean the off-chain project id @blushi ?

jsonld/credit-classes/C01-verified-carbon-standard-class.json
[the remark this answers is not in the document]
S4mmyb: awesome, thanks!

jsonld/credit-batches/C01-verified-carbon-standard.json
wgwz: Nevermind my last comment, somehow I didn’t see you had resolved it.

jsonld/credit-classes/C01-verified-carbon-standard-class.json
```diff
@@ -0,0 +1,43 @@
...
+    "schema": "http://schema.org/",
+    "regen": "http://regen.network/",
+    "xsd": "http://www.w3.org/2001/XMLSchema#"
+  },
+  "@type": "regen:CreditClass",
+  "schema:description": "",
+  "regen:sourceRegistry": {
+    "@type": "regen:Organization",
+    "schema:name": ""
+  },
```
blushi, line 12: @S4mmyb @wgwz we might also want to add the schema:url here since the UI needs it: https://www.figma.com/file/seUiQM2SPt78KJiDL0Ea8m/Basic-Ecocredits-Ledger-View?node-id=587%3A77580

jsonld/credit-classes/C01-verified-carbon-standard-class.json
```diff
@@ -0,0 +1,43 @@
...
+  "regen:sectoralScope": {
+    "@type": "schema:ItemList",
+    "schema:itemListElement": []
+  },
+  "regen:offsetGenerationMethod": {
+    "@type": "schema:ItemList",
+    "schema:itemListElement": []
+  },
+  "regen:approvedMethodologies": {
+    "@type": "schema:ItemList",
```
mhagel, line 22: @S4mmyb @blushi  In order to link to the list here (example: https://library.regen.network/v/regen-registry-credit-classes/credits-from-other-registries/verified-carbon-standard-credit-class/approved-methodologies), do you think we need to add a URL to `"regen:approvedMethodologies"`, or should each methodology have the same URL, pointing to the parent list?

Relevant UI: https://www.figma.com/file/seUiQM2SPt78KJiDL0Ea8m/Basic-Ecocredits-Ledger-View?node-id=587%3A77580

jsonld/credit-classes/C01-verified-carbon-standard-class.json
S4mmyb: Yep, it should be a list of strings. Something along the lines of `"sectoralScope": ["Agriculture Forestry and Other Land Use (AFOLU)", "Livestock and manure management"]`. I was originally using `schema:itemList` because I didn't know any better, but based on @blushi's comment we should stick with using `dash:ListShape` as this data doesn't require any ordering or numbering. This same logic applies to the `approvedMethodology` list and any other lists in the JSONLD files rn. 

As to your question on sectoral scope, sectoral scopes are used to categorize projects based on the project activity or development sector. Every registry  (verra, gold standard, cdm, etc...) has their own definitions of sectoral scope, but for this credit class we are referencing those listed in the Verified Carbon Standard Program (https://verra.org/project/vcs-program/projects-and-jnr-programs/vcs-sectoral-scopes/). While they list 15 sectoral scopes, our registry is only focused on ecological regeneration projects, hence we are only approving the final two: AFOLU & Livestock and Manure Management.

jsonld/credit-classes/C01-verified-carbon-standard-class.json
S4mmyb: Good point, thanks @blushi for catching that :)

jsonld/credit-classes/C01-verified-carbon-standard-class.json
S4mmyb: Hmm, that's a good point. I was originally thinking we could add the name, url, and version number for each methodology into the schema, but it might be nice to also include a link to the full list of approved methodologies so you can pull that into the UI.  Maybe we could also add a field before the `schema:ItemList` in the `approvedMethodologies` section to add in the gitbook link? Maybe something along the lines of: 

`
  "regen:methodologyList": {
    "@type": "schema:URL",
    "@value": ""
  },
`

Thoughts @blushi @wgwz @mhagel?

blushi: Then I think for this particular use case where we need a same property (`regen:approvedMethodologies`) to have both a list (of methologies) and some general url, we might want to consider using https://schema.org/BreadcrumbList

blushi: this would result in something like this:
```
 "regen:approvedMethodologies": {
    "@type": "schema:BreadcrumbList",
    "schema:itemListElement": [
      {
        "@type": "regen:Methodology",
        "schema:name": "methodology 1 name",
        "schema:url": {
          "@type": "schema:URL",
          "@value": "http://foo1.com"
        }
      },
      {
        "@type": "regen:Methodology",
        "schema:name": "methodology 2 name",
        "schema:url": {
          "@type": "schema:URL",
          "@value": "http://foo2.com"
        }
      }
    ],
    "schema:url": "http://bar.com"
 }
```

jsonld/credit-classes/C01-verified-carbon-standard-class.json
wgwz: I believe I've added the SHACL and corresponding JSONLD example correctly, based on what was discussed here. To be honest, I'm not very confident I've got it right but I tried to follow along with the example you provided @blushi. The commit for this change is cb74b5d9ecd9fc82ef46b18c2ce205b902eb2851.

jsonld/credit-classes/C01-verified-carbon-standard-class.json
wgwz: Ok, again I think I got this right now. But open to suggestions because I'm still feeling not too confident that I've wrapped my head around SHACL. This change is made in this commit 8d50696.

jsonld/credit-classes/C01-verified-carbon-standard-class.json
wgwz: I attempted to add the SHACL schema for this piece as well. But I'm less confident on this one, in fact I know it's wrong :-) and I'm feeling somewhat lost on how to write the SHACL for this approvedMethodologies piece, while using the BreadcrumbList. I pushed my work so far on this, it's in this commit 7f78cf0d0f610db5b2ba342abeca120e98640b27.

jsonld/credit-classes/C01-verified-carbon-standard-class.json
```diff
@@ -3,7 +3,8 @@
     "schema": "http://schema.org/",
     "regen": "http://regen.network/",
     "xsd": "http://www.w3.org/2001/XMLSchema#",
-    "dash": "http://datashapes.org/dash#"
+    "dash": "http://datashapes.org/dash#",
+    "rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#"
```
wgwz, line 7: needed to add rdf here for the graph to validate..

jsonld/credit-classes/C01-verified-carbon-standard-class.json
```diff
@@ -28,28 +29,30 @@
...
-        }
-      },
-      {
-        "@type": "regen:Methodology",
-        "schema:name": "",
-        "schema:url": {
-          "@type": "schema:URL",
-          "@value": ""
+    "schema:itemListElement": {
+      "@list": [
```
wgwz, line 33: @blushi i could remember the syntax shortcut you showed me earlier today so i used this way for now. glad to change that if you could remind me the syntax, something like "component=>list" higher up in the jsonld.

shacl/credit-classes/C01-verified-carbon-standard-class.ttl
```diff
@@ -38,9 +38,16 @@ regen:CreditClassShape a sh:NodeShape ;
...
+    sh:class schema:BreadcrumbList ;
     sh:property [
       sh:path schema:itemListElement ;
+      sh:minCount 1 ;
+      sh:property [
+        sh:path ( [ sh:zeroOrMorePath rdf:rest ] rdf:first ) ;
+        sh:node regen:MethodologyVersionShape ;
+        sh:class regen:Methodology ;
+        sh:minCount 1 ;
+      ] ;
```
wgwz, line 50: thanks for showing me how to handle this case @blushi ! i think i've got it correct now..

jsonld/credit-classes/C01-verified-carbon-standard-class.json
blushi: yes please add to "@context" the following line:
`"schema:itemListElement": { "@container": "@list" }`
see docs: https://w3c.github.io/json-ld-syntax/#example-specifying-that-a-collection-is-ordered-in-the-context

jsonld/credit-classes/C01-verified-carbon-standard-class.json
blushi: what do you mean? else what does return the shacl validation?

jsonld/credit-batches/C01-verified-carbon-standard-batch.json
```diff
@@ -0,0 +1,20 @@
...
+    "@value": ""
+  },
+  "regen:batchMonitoringReport": {
+    "@type": "schema:URL",
+    "@value": ""
+  },
+  "regen:batchVerificationReport": {
+    "@type": "schema:URL",
+    "@value": ""
+  }
```
blushi, line 19: @wgwz we're gonna need to add some `schema:name` for both of these (see https://www.figma.com/file/seUiQM2SPt78KJiDL0Ea8m?node-id=1448:95862#164614688), so it could just look like this:
```
"regen:batchVerificationReport": {
  "schema:name": "",
  "schema:url": {
    "@type": "schema:URL",
    "@value": ""
  }
}
```

jsonld/credit-classes/C01-verified-carbon-standard-class.json
wgwz: i'm not sure why i added or thought that i needed this now. at some point i thought it fixed something but i can't remember to prove/disprove. removing it for now.

jsonld/credit-batches/C01-verified-carbon-standard-batch.json
wgwz: i've made use of the NameUrlShape for these reports now, via a71a0be

README.md
```diff
@@ -9,3 +9,7 @@ This repository contains:
 ## SHACL Graphs
 
 The `shacl` folder contains SHACL schemas for validating data (for example, project related data), using [Turtle](https://www.w3.org/TR/turtle/) or [JSON-LD](https://json-ld.org/).
+
+## JSON-LD Examples
+
+The `jsonld` folder contains examples of JSON-LD data that can be directly copied/pasted and filled in (filling in empty strings and replacing `0` with appropriate numbers). TODO: Corresponding SHACL graphs can be found in the `shacl` folder.
```
blushi, line 15:
```suggestion
The `jsonld` folder contains examples of JSON-LD data that can be directly copied/pasted and filled in (filling in empty strings and replacing `0` with appropriate numbers). Corresponding SHACL graphs can be found in the `shacl` folder.
```

jsonld/credit-classes/C01-verified-carbon-standard-class.json
```diff
@@ -0,0 +1,56 @@
...
+    "schema:url": {
+      "@type": "schema:URL",
+      "@value": ""
+    }
+  },
+  "regen:sectoralScope": [
+    {
+      "@type": "xsd:string",
+      "@value": ""
+    }
```
blushi, line 22: no need to specify the type here, just a string is enough and should work fine with the validation
```suggestion
""
```

jsonld/credit-classes/C01-verified-carbon-standard-class.json
```diff
@@ -0,0 +1,56 @@
...
+  ],
+  "regen:offsetGenerationMethod": [
+    {
+      "@type": "xsd:string",
+      "@value": ""
+    }
```
blushi, line 28: ditto

jsonld/credit-classes/C01-verified-carbon-standard-class.json
```diff
@@ -0,0 +1,56 @@
+{
+  "@context": {
+    "schema": "http://schema.org/",
+    "regen": "http://regen.network/",
+    "xsd": "http://www.w3.org/2001/XMLSchema#",
+    "dash": "http://datashapes.org/dash#",
+    "schema:itemListElement": { "@container": "@list" }
```
blushi, line 7: you also need to add the same for `regen:sectoralScope` and `regen:offsetGenerationMethod` or the shacl validation will always fail

shacl/credit-classes/C01-verified-carbon-standard-class.ttl
```diff
@@ -0,0 +1,66 @@
...
+      sh:datatype schema:URL ;
+    ] ;
+    sh:minCount 1 ;
+    sh:maxCount 1 ;
+  ] ;
+  sh:property [
+    sh:path regen:verificationMethod ;
+    sh:minCount 1 ;
+    sh:maxCount 1 ;
+    sh:minLength 1 ;
```
blushi, line 63: why adding a minLength here and not for the other string values?

blushi (commented):
Looks good, some small remaining comments

shacl/credit-classes/C01-verified-carbon-standard-class.ttl
wgwz: i'll go ahead and add these to the other string types as well, it's just a oversight, sorry about that.

jsonld/credit-classes/C01-verified-carbon-standard-class.json
wgwz: done! added these via 7b6831fb075491025c11b516bc7bf762289f5691, thanks for pointing this out. i tested it in the playground and indeed it was not validating without this.

jsonld/credit-classes/C01-verified-carbon-standard-class.json
wgwz: got it, thanks for pointing this out..

jsonld/projects/C01-verified-carbon-standard-project.json
```diff
@@ -0,0 +1,56 @@
...
+  "@type": "regen:Project",
+  "regen:vcsProjectId": {
+    "@type": "xsd:unsignedInt",
+    "@value": 0
+  },
+  "regen:vcsProjectPage": {
+    "@type": "schema:URL",
+    "@value": ""
+  },
+  "regen:projectProponent": {
```
mhagel, line 18: @S4mmyb I think we need some more fields here to match the comps:

```
{
  name: string;
  location?: string;
  image?: string | null;
  description?: string;
  link?: string;
}
```

mhagel, line 18: @S4mmyb I think we need some more fields here to match the comps:

```
{
  name: string;
  location?: string;
  image?: string | null;
  description?: string;
  link?: string;
}
```

blushi: @wgwz `projectProponent` should be changed to `projectDeveloper` too per the discussion on slack: https://regen-network.slack.com/archives/C01LX9E8QN8/p1647452208424579

blushi: we could also just hide the location and description (these looks like there are already optional anyway in the typescript interface) and use some placeholder avatar in place of the image @mhagel

S4mmyb: Hey @mhagel! Thanks for the patience on the response here. After talking with @clevinson I think that we don't necessarily need to be pulling all this information from the project metadata. I think we could pull image and link in from the regular database, correct? As it relates to location and link, I think we can remove those, but I can add description into the PD metadata as follows: 

`"regen:projectDeveloper": {
     "schema:name": "",
     "schema:description": ""
}
`

S4mmyb: Hey @mhagel, thanks for the patience on my response. I actually don't think we need to store all of that in the project metadata, some of it we can store in our registry database. With that in mind, I think we could put the images in the database and actually drop location and link all together. As it relates to description, we can add it to the metadata here as follows: 
`"regen:projectProponent": {
    "schema:name": "",
    "schema:description": ""
}

wgwz: Hi all :wave:  I've made some changes based on our convo's, glad to make further changes if needed.

1. renamed projectProponent to projectDeveloper via 7568fd2
2. added a description field to projectDeveloper via 8a9bedf

seconding Mark's last question as well. i was unsure if i could reuse that `OrganizationDisplayShape`, so that's why i created a new one in `shacl/projects/C01-verified-carbon-standard-project.ttl`

jsonld/projects/C01-verified-carbon-standard-project.json
```diff
@@ -0,0 +1,56 @@
+{
+  "@context": {
+    "schema": "http://schema.org/",
+    "regen": "http://regen.network/",
+    "qudt": "http://qudt.org/schema/qudt",
+    "unit": "http://qudt.org/2.1/vocab/unit",
+    "xsd": "http://www.w3.org/2001/XMLSchema#"
+  },
+  "@type": "regen:Project",
```
mhagel, line 9: Do we want a project name? @S4mmyb 
```suggestion
  "@type": "regen:Project",
  "schema:name": ""
```

mhagel: We have a note in the code to show the on-chain projectId if there is no name. I could just do that.

jsonld/projects/C01-verified-carbon-standard-project.json
blushi: Mark's proposal sounds good to me 
we should probably also add: `"regen:showOnProjectPage": true` then

jsonld/projects/C01-verified-carbon-standard-project.json
blushi: I think it would make sense to add the project name indeed
although we'll only be able to add the on-chain project ID once we have projects on-chain (probably not before a few months)

jsonld/projects/C01-verified-carbon-standard-project.json
wgwz: I've made these changes via 3a565abd0366acea081b2de4b5a9ba24a14e2cce

jsonld/projects/C01-verified-carbon-standard-project.json
wgwz: I made this addition via fb5be86b49ff521f75a1aa48a610c5a70e0d55ca

jsonld/projects/C01-verified-carbon-standard-project.json
```diff
@@ -0,0 +1,59 @@
...
+  "regen:projectSize": {
+    "qudt:unit": {
+      "@type": "qudt:Unit",
+      "@value": "unit:HA"
+    },
+    "qudt:numericValue": {
+      "@type": "xsd:double",
+      "@value": 0
+    }
+  },
```
mhagel, line 50: @S4mmyb We may need location here. It is in the Project Page [standard](https://github.com/regen-network/regen-registry-standards/blob/3a565abd0366acea081b2de4b5a9ba24a14e2cce/shacl/projects/project-page.ttl#L48). 

@blushi Do we need to include location in the standard, or can we just consume it in the front end if it is available, like we do now with `ProjectPageShape`?

```suggestion
  },
  "schema:location": {
      ...location fields
  },
```

mhagel: for reference, this is what Wilmot looks like

```
  "http://schema.org/location":
  {
      "id": "locality.6875563885184200",
      "bbox":
      [
          152.281925823569,
          -30.4188967883064,
          152.53051845185,
          -30.1967091580822
      ],
      "text": "Hernani",
      "type": "Feature",
      "center":
      [
          152.4184,
          -30.2986
      ],
      "context":
      [
          {
              "id": "postcode.8076814917761920",
              "text": "2453"
          },
          {
              "id": "region.19582497656831280",
              "text": "New South Wales",
              "wikidata": "Q3224",
              "short_code": "AU-NSW"
          },
          {
              "id": "country.9665923154346070",
              "text": "Australia",
              "wikidata": "Q408",
              "short_code": "au"
          }
      ],
      "@context":
      {
          "type": "@type",
          "@vocab": "https://purl.org/geojson/vocab#",
          "coordinates":
          {
              "@container": "@list"
          }
      },
      "geometry":
      {
          "type": "Point",
          "coordinates":
          [
              152.4184,
              -30.2986
          ]
      },
      "relevance": 1,
      "place_name": "Hernani, New South Wales, Australia",
      "place_type":
      [
          "locality"
      ],
      "properties": {}
  },
```

mhagel: maybe also `schema:boundaries` if we want to have geojson maps

jsonld/projects/C01-verified-carbon-standard-project.json
```diff
@@ -0,0 +1,59 @@
...
+  "regen:vcsProjectType": "",
+  "regen:projectActivity": {
+   "schema:name": "",
+    "schema:url": {
+      "@type": "schema:URL",
+      "@value": ""
+    }
+  },
+  "regen:offsetGenerationMethod": "",
```
mhagel, line 33: We may also need `regen:landStory`
@blushi @S4mmyb

jsonld/projects/C01-verified-carbon-standard-project.json
blushi: see my comment on slack https://regen-network.slack.com/archives/C01LX9E8QN8/p1647934515526819?thread_ts=1647612024.927119&cid=C01LX9E8QN8, we could use either `schema:location` or `schema:boundaries` for the map component

jsonld/credit-classes/C01-verified-carbon-standard-class.json
```diff
@@ -0,0 +1,52 @@
+{
```
wgwz, line 1: <ins>Todo</ins>: Update the `regen:Methodology` schema to be consistent with [the credit class @S4mmyb put on chain][1] and also make sure that the other metadata here is consistent with regen-registry-ops repo.

[1]: https://github.com/regen-network/regen-registry-ops/blob/sam/C01-metadata/C01/credit-class-metadata/C01-verified-carbon-standard-credit-class.json

wgwz: For reference: https://github.com/regen-network/rnd-tx-ops/pull/8#issuecomment-1074853566

jsonld/projects/C01-verified-carbon-standard-project.json
wgwz: I added this via 88b0f29. But I just copied the `regen:landStory` property from the `project-page.ttl` schema and added it the C01 project schema. I'm not sure if that's how we wanted to do this or if there is a better way.

jsonld/projects/C01-verified-carbon-standard-project.json
```diff
@@ -0,0 +1,59 @@
...
+    }
+  },
+  "regen:projectStartDate": {
+    "@type": "xsd:date",
+    "@value": ""
+  },
+  "regen:projectEndDate": {
+    "@type": "xsd:date",
+    "@value": ""
+  }
```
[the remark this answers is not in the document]
wgwz: i've added the additional certifications field via 84166c8 and eab8ca1. i've made this field so that it can only have one single value. and the url is marked as optional in the schema. c.c. @blushi please review these when you can, just in case i misunderstood what we needed

jsonld/projects/C01-verified-carbon-standard-project.json
wgwz: so just wanting to confirm that the todo item here is we need to add the geojson `schema:location` field, right? or is there something more to it?

jsonld/credit-classes/C01-verified-carbon-standard-class.json
wgwz: I made these updates via dff0009 and 794e934. I think we are now in sync with @S4mmyb's credit class in regen-registry-ops.

One question I had in relation to these changes is if I should mark the newly added `schema:identifier` and `schema:version` fields on our `regen:MethodologyVersionShape` as required or not, I'm guessing we probably do but I was worried about backwards compatibility (which may not matter). c.c. @blushi

jsonld/projects/C01-verified-carbon-standard-project.json
blushi: fine by me for now

jsonld/projects/C01-verified-carbon-standard-project.json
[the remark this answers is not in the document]
blushi: LGTM

jsonld/projects/C01-verified-carbon-standard-project.json
blushi: yes that's all we need to add

shacl/projects/C01-verified-carbon-standard-project.ttl
```diff
@@ -0,0 +1,135 @@
...
+  sh:property [
+    sh:path schema:url ;
+    sh:minCount 1 ;
+    sh:maxCount 1 ;
+    sh:datatype schema:URL ;
+  ] ;
+.
+
+regen:NameOptionalUrlShape a sh:NodeShape ;
+  sh:targetClass regen:NameOptionalUrl;
```
blushi, line 109: this is not necessarily needed here

jsonld/credit-classes/C01-verified-carbon-standard-class.json
blushi: > One question I had in relation to these changes is if I should mark the newly added `schema:identifier` and `schema:version` fields on our `regen:MethodologyVersionShape` as required or not, I'm guessing we probably do but I was worried about backwards compatibility (which may not matter). c.c. @blushi

I guess this is more a question for @S4mmyb

jsonld/projects/C01-verified-carbon-standard-project.json
```diff
@@ -0,0 +1,67 @@
+{
+  "@context": {
+    "schema": "http://schema.org/",
+    "regen": "http://regen.network/",
+    "qudt": "http://qudt.org/schema/qudt",
+    "unit": "http://qudt.org/2.1/vocab/unit",
```
blushi, line 6:
```suggestion
    "qudt": "http://qudt.org/schema/qudt/",
    "unit": "http://qudt.org/2.1/vocab/unit/",
```

shacl/projects/C01-verified-carbon-standard-project.ttl
```diff
@@ -0,0 +1,135 @@
+@prefix schema: <http://schema.org/> .
+@prefix sh: <http://www.w3.org/ns/shacl#> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+@prefix regen: <http://regen.network/> .
+@prefix qudt: <http://qudt.org/schema/qudt> .
```
blushi, line 5:
```suggestion
@prefix qudt: <http://qudt.org/schema/qudt/> .
```

shacl/projects/project-page.ttl
```diff
@@ -7,7 +7,7 @@
 @prefix regen: <http://regen.network/> .
 @prefix geojson: <https://purl.org/geojson/vocab#> .
 @prefix qudt_unit: <http://qudt.org/1.1/vocab/unit#> .
-@prefix qudt: <http://qudt.org/1.1/schema/qudt#> .
+@prefix qudt: <http://qudt.org/schema/qudt> .
```
blushi, line 10:
```suggestion
@prefix qudt: <http://qudt.org/schema/qudt/> .
```

jsonld/projects/C01-verified-carbon-standard-project.json
wgwz: done! i'm going to add new comments on these additions, i think it probably needs some feedback.

jsonld/projects/C01-verified-carbon-standard-project.json
```diff
@@ -0,0 +1,83 @@
...
+  },
+  "schema:location": {
+    "@type": "geojson:Feature",
+    "geojson:geometry": {
+      "@type": "geojson:Point",
+      "geojson:coordinates": [0.1, 0.2]
+    },
+    "geojson:properties": {},
+    "geojson:id": ""
+  }
```
wgwz, line 82: here's the bare minimum for the `schema:location` addition, i was unsure of what fields we needed to have here. i know have this example below, but i'm unsure of what to do for the fields that are here, that don't seem to be a part of [the geojson-ld spec](https://www.rfc-editor.org/rfc/rfc7946.html) i.e. `place_name`.

```
{
  "http://schema.org/location": {
    "id": "locality.6875563885184200",
    "bbox": [
      152.281925823569,
      -30.4188967883064,
      152.53051845185,
      -30.1967091580822
    ],
    "text": "Hernani",
    "type": "Feature",
    "center": [
      152.4184,
      -30.2986
    ],
    "context": [
      {
        "id": "postcode.8076814917761920",
        "text": "2453"
      },
      {
        "id": "region.19582497656831280",
        "text": "New South Wales",
        "wikidata": "Q3224",
        "short_code": "AU-NSW"
      },
      {
        "id": "country.9665923154346070",
        "text": "Australia",
        "wikidata": "Q408",
        "short_code": "au"
      }
    ],
    "@context": {
      "type": "@type",
      "@vocab": "https://purl.org/geojson/vocab#",
      "coordinates": {
        "@container": "@list"
      }
    },
    "geometry": {
      "type": "Point",
      "coordinates": [
        152.4184,
        -30.2986
      ]
    },
    "relevance": 1,
    "place_name": "Hernani, New South Wales, Australia",
    "place_type": [
      "locality"
    ],
    "properties": {}
  }
}
```

shacl/projects/C01-verified-carbon-standard-project.ttl
```diff
@@ -0,0 +1,141 @@
...
+    sh:path regen:additionalCertification ;
+    sh:node regen:NameOptionalUrlShape ;
+    sh:maxCount 1 ;
+  ] ;
+  sh:property [
+    sh:class geojson:Feature ;
+    sh:path schema:location ;
+    sh:minCount 1 ;
+    sh:maxCount 1 ;
+  ] ;
```
wgwz, line 95: Here's the bare minimum for `schema:location` on the SHACL side. I was unsure how specific we need or want to be here. Like, i.e. do we need to specify that the `bbox` needs to be a list of four `xsd:double`'s?

jsonld/projects/C01-verified-carbon-standard-project.json
blushi: I think we should just leave an empty object in the json template since the full geojson data should be generated through some geo tool like a geocoder (https://docs.mapbox.com/playground/geocoding/)

shacl/projects/C01-verified-carbon-standard-project.ttl
blushi: please see existing project-page.ttl file, there's already a property for such `schema:location` in there
what's important there is in particular `sh:class geojson:Feature ;`

jsonld/projects/C01-verified-carbon-standard-project.json
wgwz: done! via f5f008914c4d72971c7fe49e493c6d0e662bb570

shacl/projects/C01-verified-carbon-standard-project.ttl
wgwz: yup, got that down in this schema already, think we're good.

jsonld/credit-classes/C01-verified-carbon-standard-class.json
S4mmyb: @wgwz I think we do want these to be required. Most registry programs (Verra, CDM, GS, etc...) use identifiers and versions and this will be something we will incorporate into the registry program as well.

jsonld/credit-classes/C01-verified-carbon-standard-class.json
```diff
@@ -0,0 +1,73 @@
...
+  "@context": {
+    "schema": "http://schema.org/",
+    "regen": "http://regen.network/",
+    "xsd": "http://www.w3.org/2001/XMLSchema#",
+    "dash": "http://datashapes.org/dash#",
```
S4mmyb, line 6: @wgwz we don't use dash here so we can remove it 
```suggestion
```

jsonld/credit-classes/C01-verified-carbon-standard-class.json
```diff
@@ -0,0 +1,73 @@
...
+  },
+  "regen:sourceRegistry": {
+    "schema:name": "",
+    "schema:url": {
+      "@type": "schema:URL",
+      "@value": ""
+    }
+  },
+  "regen:sectoralScope": [
+    ""
```
S4mmyb, line 26: @wgwz do we want to add a type template here so folks know what should be put in? 

```suggestion
    {
      "@type": "xsd:string",
      "@value": ""
    }
```

jsonld/credit-classes/C01-verified-carbon-standard-class.json
```diff
@@ -0,0 +1,73 @@
...
+  ],
+  "regen:offsetGenerationMethod": [
+    ""
```
S4mmyb, line 29: Same as above... 

```suggestion
    {
      "@type": "xsd:string",
      "@value": "Emission Reductions"
    }
```

S4mmyb (approved):
Hey @wgwz, thanks for the diligence on this PR! I made a few minor suggestions on the credit class JSONLD file and left reviewing the Project & Batch metadata files for our future PR. Looks great to me!

jsonld/credit-classes/C01-verified-carbon-standard-class.json
wgwz: So at some point @S4mmyb I thought we needed to have that explicit typing for the `xsd:string` but @blushi clarified that in this case we don't need to so explicit. Honestly, I'm not 100% clear on the reasons why the explicit typing seemed to be required for certain cases in order for validation to happen. I know we had some of those. I attribute it to me just  eing a noob with RDF ;-)

In any case, I think Marie wants this to just be an empty string. So I think it's OK as is.

jsonld/credit-classes/C01-verified-carbon-standard-class.json
wgwz: indeed! good catch

jsonld/credit-classes/C01-verified-carbon-standard-class.json
wgwz: Ok, I've marked these two fields as required for now (via 5b518a2)

jsonld/credit-classes/C01-verified-carbon-standard-class.json
wgwz: I should have said, that so far, in these JSONLD templates, when we have an example that's a string, we seem to be favoring just showing an empty string, rather than some example/fake value. so `""` over `"Foo Inc."` So it's more that we're just trying to be consistent with that (I guess..)

wgwz: Also, I guess, that the way you've specified the string data:

```
[
    {
      "@type": "xsd:string",
      "@value": ""
    }
]
```

Is perfectly equivalent to `[""]`. It's two different ways to represent this, I think it means exactly the same thing in RDF perspective.

wgwz: Marking resolved :-)

S4mmyb: awesome, thanks!

blushi: On https://regen-registry-server.herokuapp.com/metadata-graph/regen:13toVgqN4ESXgLNQZNz7iGPNuwr7HghWxNCk4W5JrsaX2NwXfNJVg9K.rdf, the C01 credit class metadata was still using xsd:string format, I've updated it to just use strings so we keep things consistent (the generated IRI is unchanged since the data is exactly the same as @wgwz pointed out).

blushi: FYI @S4mmyb I've updated the credit class metadata to just use strings for the CO1 class in the db

blushi (approved):

jsonld/credit-classes/C01-verified-carbon-standard-class.json
blushi: sure! thanks

PR #16, opened 2022-05-09
Update VCS project metadata SHACL graph with `sh:group`s

blushi:
## Description

Closes: https://github.com/regen-network/regen-registry/issues/909

Needed for front-end validation of project forms.
This is based on https://www.figma.com/file/TyhFaXqko9F8OvLEE4Zomg/Issuer-Create%2FEdit-Projects%2C-Credit-Class%2C-Batches?node-id=2%3A4449

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/shacl/credit-batches/C01-verified-carbon-standard-batch.ttl
+++ b/shacl/credit-batches/C01-verified-carbon-standard-batch.ttl
@@ -5,7 +5,7 @@
 @prefix dash: <http://datashapes.org/dash#> .
 @prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
 
-regen:CreditBatchShape a sh:NodeShape ;
+regen:C01-CreditBatchShape a sh:NodeShape ;
   sh:targetClass regen:CreditBatch ;
   sh:property [
     sh:path regen:vcsRetirementSerialNumber ;
--- a/shacl/credit-batches/C01-verified-carbon-standard-batch.ttl
+++ b/shacl/credit-batches/C01-verified-carbon-standard-batch.ttl
@@ -22,9 +22,7 @@ regen:C01-CreditBatchShape a sh:NodeShape ;
     sh:property [
       sh:path ( [ sh:zeroOrMorePath rdf:rest ] rdf:first ) ;
       sh:node regen:NameOptionalUrlShape ;
-      sh:minCount 1 ;
     ] ;
-    sh:minCount 1 ;
     sh:maxCount 1 ;
   ] ;
 .
--- a/shacl/credit-batches/C01-verified-carbon-standard-batch.ttl
+++ b/shacl/credit-batches/C01-verified-carbon-standard-batch.ttl
@@ -5,8 +5,10 @@
 @prefix dash: <http://datashapes.org/dash#> .
 @prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
 
+regen:regen:C01-CreditBatch rdfs:subClassOf regen:CreditBatch .
+
 regen:C01-CreditBatchShape a sh:NodeShape ;
-  sh:targetClass regen:CreditBatch ;
+  sh:targetClass regen:C01-CreditBatch ;
   sh:property [
     sh:path regen:vcsRetirementSerialNumber ;
     sh:minCount 1 ;
--- a/shacl/credit-batches/C01-verified-carbon-standard-batch.ttl
+++ b/shacl/credit-batches/C01-verified-carbon-standard-batch.ttl
@@ -26,3 +26,18 @@ regen:CreditBatchShape a sh:NodeShape ;
     sh:maxCount 1 ;
   ] ;
 .
+
+regen:NameOptionalUrlShape a sh:NodeShape ;
+  sh:property [
+    sh:path schema:name ;
+    sh:minCount 1 ;
+    sh:maxCount 1 ;
+    sh:minLength 1 ;
+    sh:datatype xsd:string ;
+  ] ;
+  sh:property [
+    sh:path schema:url ;
+    sh:maxCount 1 ;
+    sh:datatype schema:URL ;
+  ] ;
+.
--- a/shacl/projects/C01-verified-carbon-standard-project.ttl
+++ b/shacl/projects/C01-verified-carbon-standard-project.ttl
@@ -85,7 +85,7 @@ regen:C01-ProjectShape a sh:NodeShape ;
   ] ;
   sh:property [
     sh:path regen:vcsProjectId ;
-    sh:datatype xsd:unsignedInt ;
+    sh:datatype xsd:integer ;
     sh:minCount 1 ;
     sh:maxCount 1 ;
     sh:group regen:ProjectPageMetadataGroup ;
--- a/shacl/projects/C01-verified-carbon-standard-project.ttl
+++ b/shacl/projects/C01-verified-carbon-standard-project.ttl
@@ -9,7 +9,7 @@
 @prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
 @prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
 
-regen:ProjectShape a sh:NodeShape ;
+regen:C01-ProjectShape a sh:NodeShape ;
   sh:targetClass regen:Project ;
   sh:property [
     sh:path schema:name ;
--- a/shacl/projects/C01-verified-carbon-standard-project.ttl
+++ b/shacl/projects/C01-verified-carbon-standard-project.ttl
@@ -228,7 +228,7 @@ regen:ProjectPageDescriptionGroup
 
 regen:ProjectPageMediaGroup
 	a sh:PropertyGroup ;
-	rdfs:label "Photos" ;
+	rdfs:label "Media" ;
 	sh:order "5"^^xsd:decimal .
 
 regen:ProjectPageMetadataGroup
--- a/shacl/projects/C01-verified-carbon-standard-project.ttl
+++ b/shacl/projects/C01-verified-carbon-standard-project.ttl
@@ -9,8 +9,10 @@
 @prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
 @prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
 
+regen:C01-Project rdfs:subClassOf regen:Project .
+
 regen:C01-ProjectShape a sh:NodeShape ;
-  sh:targetClass regen:Project ;
+  sh:targetClass regen:C01-Project ;
   sh:property [
     sh:path schema:name ;
     sh:minCount 1 ;
--- a/shacl/projects/C01-verified-carbon-standard-project.ttl
+++ b/shacl/projects/C01-verified-carbon-standard-project.ttl
@@ -21,7 +21,7 @@ regen:ProjectShape a sh:NodeShape ;
   ] ;
   sh:property [
     sh:path regen:projectSize ;
-    sh:node regen:SizeShape ;
+    sh:node regen:ProjectSizeShape ;
     sh:minCount 1 ;
     sh:maxCount 1 ;
     sh:group regen:ProjectPageBasicInfoGroup ;
@@ -146,7 +146,7 @@ regen:NameUrlShape a sh:NodeShape ;
   ] ;
 .
 
-regen:SizeShape a sh:NodeShape ;
+regen:ProjectSizeShape a sh:NodeShape ;
   sh:property [
     sh:path qudt:unit ;
     sh:datatype qudt:Unit ;
--- a/shacl/projects/C01-verified-carbon-standard-project.ttl
+++ b/shacl/projects/C01-verified-carbon-standard-project.ttl
@@ -42,9 +42,6 @@ regen:ProjectShape a sh:NodeShape ;
   ] ;
   sh:property [
     sh:path schema:description ;
-    sh:minCount 1 ;
-    sh:maxCount 1 ;
-    sh:minLength 1 ;
     sh:maxLength 600 ;
     sh:datatype xsd:string ;
     sh:group regen:ProjectPageDescriptionGroup ;
@@ -52,8 +49,7 @@ regen:ProjectShape a sh:NodeShape ;
   sh:property [
     sh:path regen:previewPhoto ;
     sh:datatype schema:URL ;
-    sh:name "Preview photo" ;
-    sh:group regen:ProjectPagePhotosGroup ;
+    sh:group regen:ProjectPageMediaGroup ;
   ] ;
   sh:property [
     sh:path regen:galleryPhotos ;
@@ -61,9 +57,14 @@ regen:ProjectShape a sh:NodeShape ;
     sh:property [
       sh:path ( [ sh:zeroOrMorePath rdf:rest ] rdf:first ) ;
       sh:datatype schema:URL ;
-      sh:group regen:ProjectPagePhotosGroup ;
+      sh:group regen:ProjectPageMediaGroup ;
     ] ;
-    sh:group regen:ProjectPagePhotosGroup ;
+    sh:group regen:ProjectPageMediaGroup ;
+  ] ;
+  sh:property [
+    sh:path regen:videoURL ;
+    sh:datatype schema:URL ;
+    sh:group regen:ProjectPageMediaGroup ;
   ] ;
   sh:property [
     sh:path regen:offsetGenerationMethod ;
@@ -166,10 +167,11 @@ regen:SizeShape a sh:NodeShape ;
 
 regen:RoleShape
   a sh:NodeShape ;
-  sh:targetClass regen:Individual ;
-  sh:targetClass regen:IndividualDisplay ;
-  sh:targetClass regen:Organization ;
-  sh:targetClass regen:OrganizationDisplay ;
+  sh:targetClass
+    regen:Individual ,
+    regen:IndividualDisplay ,
+    regen:Organization ,
+    regen:OrganizationDisplay ;
   sh:property [
     sh:path regen:showOnProjectPage ;
     sh:datatype xsd:boolean ;
@@ -192,10 +194,12 @@ regen:RoleShape
   sh:property [
     sh:path schema:description ;
     sh:datatype xsd:string ;
-    sh:minCount 1 ;
-    sh:maxCount 1 ;
-    sh:minLength 1 ;
     sh:maxLength 160 ;
+    sh:group regen:ProjectPageRolesGroup ;
+  ] ;
+  sh:property [
+    sh:path regen:address ;
+    sh:datatype xsd:string ;
     sh:group regen:ProjectPageRolesGroup ;
   ] ;
 .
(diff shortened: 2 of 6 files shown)


shacl/projects/project-page.ttl
```diff
@@ -273,8 +273,8 @@ regen:ProjectSizeShape
   ] ;
   sh:property [
     sh:path qudt:unit ;
-    sh:datatype qudt:unit ;
-    sh:in ( "http://qudt.org/1.1/vocab/unit#HA"^^qudt:unit "http://qudt.org/1.1/vocab/unit#AC"^^qudt:unit ) ;
+    sh:datatype qudt:Unit ;
+    sh:in ( "unit:HA"^^qudt:Unit "unit:AC"^^qudt:Unit ) ;
```
blushi, line 277: also fixed the @type value in the project metadata on the staging and production db, we were using qudt:unit (which is a property name) instead of qudt:Unit (which is the type name)

shacl/projects/C01-verified-carbon-standard-project.ttl
```diff
@@ -3,87 +3,126 @@
...
 @prefix regen: <http://regen.network/> .
 @prefix qudt: <http://qudt.org/schema/qudt/> .
+@prefix unit: <http://qudt.org/vocab/unit/> .
 @prefix geojson: <https://purl.org/geojson/vocab#> .
+@prefix dash: <http://datashapes.org/dash#> .
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
 
-regen:ProjectShape a sh:NodeShape ;
+regen:C01-ProjectShape a sh:NodeShape ;
```
blushi, line 12: I was thinking we could add the on-chain credit class id as prefix for the shape name.
This way we can use this iri to store this SHACL graph in the `shacl_graph` table and query a project shape based on the credit class id, as specified in the last part of https://github.com/regen-network/regen-registry/issues/909

wgwz: how would we make the query from the frontend? graphql? i guess it's pretty much "query shacl_graph by URI". seems like that should be possible from graphql...

wgwz: i do wonder about how well this approach would work for other types of credit classes. it's kind of a trivial comment but i wonder about how what the range of possibilities is for the credit class id. right now we just have C01 but maybe in the future we'll have water and biodiversity. would we try to seek a similar id for those "W01" or "BD01"? i'm just thinking if it gets more verbose than that, maybe we want to include as a property in the shacl graph?

shacl/projects/C01-verified-carbon-standard-project.ttl
```diff
@@ -3,87 +3,126 @@
...
   sh:targetClass regen:Project ;
```
wgwz, line 13: do we need to modify this to be `C01Project`?

shacl/projects/C01-verified-carbon-standard-project.ttl
```diff
@@ -95,39 +134,104 @@ regen:NameUrlShape a sh:NodeShape ;
...
+    sh:maxCount 1 ;
+    sh:group regen:ProjectPageBasicInfoGroup ;
   ] ;
 .
 
-regen:SizeShape a sh:NodeShape ;
-  sh:targetClass regen:Size ;
+regen:RoleShape
+  a sh:NodeShape ;
+  sh:targetClass
```
wgwz, line 170: does the fact that we have four targetClass's here mean that the following `@type` property in C01 project json could take on any of those four values? 

 https://github.com/regen-network/regen-registry-standards/blob/709a8a5382748a8dbdda761bccc1f8e4f8028d34/jsonld/projects/C01-verified-carbon-standard-project.json#L21

shacl/projects/C01-verified-carbon-standard-project.ttl
```diff
@@ -3,87 +3,126 @@
...
+    sh:maxLength 600 ;
     sh:datatype xsd:string ;
+    sh:group regen:ProjectPageDescriptionGroup ;
   ] ;
   sh:property [
-    sh:path regen:projectDeveloper ;
-    sh:node regen:OrganizationDisplay ;
-    sh:minCount 1 ;
-    sh:maxCount 1 ;
+    sh:path regen:previewPhoto ;
```
wgwz, line 50: apologies if this is getting repetitive but should we also add this to the C01 project json? i'll stop commenting on these types of issues for now and wait for your feedback.

shacl/projects/C01-verified-carbon-standard-project.ttl
blushi: yes we already do from such a graphql query for the old project page form from the front end: https://github.com/regen-network/regen-web/blob/master/web-registry/src/graphql/ShaclGraphByUri.graphql

blushi: Re: your last comment, credit class ids are supposed to be short (credit type abbreviation + sequence) so I think it should be ok.
We could also specify in the graph itself the credit class but that wouldn't be helpful for querying from the front-end.

shacl/projects/C01-verified-carbon-standard-project.ttl
```diff
@@ -95,39 +134,104 @@ regen:NameUrlShape a sh:NodeShape ;
...
-    sh:datatype xsd:double ;
+    sh:path schema:name ;
+    sh:datatype xsd:string ;
+    sh:minCount 1 ;
+    sh:maxCount 1 ;
+    sh:minLength 1 ;
+    sh:group regen:ProjectPageRolesGroup ;
+  ] ;
+  sh:property [
+    sh:path schema:image ;
```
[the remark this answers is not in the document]
blushi: we could do that, although as you said, this field is optional... maybe let's ask @S4mmyb what he thinks about it.

the purpose of this PR was only to come up with a proper C01 project SHACL graph in order to validate the new project forms, which is why I didn't modify the JSON template yet.

shacl/projects/C01-verified-carbon-standard-project.ttl
blushi: yes or could also be `regen:Organization` and `regen:OrganizationDisplay`
we have for different types (individual and org, normal and display) because the original project forms included one form for the entity basic info and another one for the entity display info, in case it should be displayed on the project page. Additionally the info for an individual user and an org were different.
now we just ask the same info for both and display the entity by default so I thought we could just re-use all those types.

shacl/projects/C01-verified-carbon-standard-project.ttl
blushi: I was wondering about this
I think that we could actually use `regen:C01-Project` for consistency, we would need to update the @type of this project in particular: https://github.com/regen-network/regen-registry-ops/blob/cb68947a7b4416776c0fc9a8289a709f6ab53d07/C01/project-metadata/C01-ProjectId934-metadata.json

additionally we could also define that `regen:C01-Project` is a subclass of `regen:Project` (see Person/Doctor example in https://www.w3.org/TR/shacl/#targetClass): `regen:C01-Project rdfs:subClassOf regen:Project .` which brings some additional context, wdyt?

blushi: the other thing is that ideally we would do the same for credit batches as well, ie having:
```
regen:C01-CreditBatchShape a sh:NodeShape ;
  sh:targetClass regen:C01-CreditBatch ;
```
but the thing is that we've already issued some credit batches on-chain using `regen:CreditBatch` as @type and those can't be updated...

wgwz: > additionally we could also define that regen:C01-Project is a subclass of regen:Project (see Person/Doctor example in https://www.w3.org/TR/shacl/#targetClass): regen:C01-Project rdfs:subClassOf regen:Project . which brings some additional context, wdyt?

conceptually this makes a lot of sense. and i think that could be helpful for us when we start to write the shacl schema's for new credit types (deduping things would likely be easier).

wgwz: > but the thing is that we've already issued some credit batches on-chain using regen:CreditBatch as @type and those can't be updated...

could you remind me why those can't be updated?

but that's unfortunate. maybe we just have to accept that `Project` means `C01-Project`. And maybe in the future, i.e. `regen:C02-Project` is a subclass of `regen:BaseProject` (`Project` could also be a subclass of `regen:BaseProject`)

shacl/projects/C01-verified-carbon-standard-project.ttl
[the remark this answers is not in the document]
wgwz: I see, that makes sense. thanks for clarifying.

shacl/projects/C01-verified-carbon-standard-project.ttl
blushi: agreed but as I said:

> he thing is that we've already issued some credit batches on-chain using `regen:CreditBatch` as `@type` and those can't be updated... 

which would result in inconsistencies between existing credit batches metadata and future ones...

shacl/projects/C01-verified-carbon-standard-project.ttl
blushi: indeed but I still think we should have a different shape name anyway
it doesn't make sense to define multiple shapes with the same name with different properties and constraints.

shacl/projects/C01-verified-carbon-standard-project.ttl
blushi: > could you remind me why those can't be updated?

credit batches are not updatable on chain

shacl/projects/C01-verified-carbon-standard-project.ttl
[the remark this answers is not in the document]
S4mmyb: This is for the organization image, correct?

S4mmyb (changes requested):
Hey @blushi @wgwz, thanks for the patience on my responses as I get back from the conference. I spent some time with @clevinson thinking through how we're structuring metadata for project pages and here are our thoughts. 

Generally speaking, the only information we should be storing in our on-chain metadata IRIs are information and relationships we want to query against. Right now the information and outlined in the credit class and credit batch seems pretty straight forward as most users will want to know this like, which methodology is used by a credit class or what the retirement serial number is for a certain batch. For projects it should be similar, and we should only store info want to query against, like the size or location, methodology used, start & end date, etc... 

With that in mind, things like images and videos should probably not be included in the metadata IRIs we store on chain. Obviously this adds in some complexity in the way we are currently structuring our SHACL schemas and building our project pages as we've blended much of the info we store off-chain, such as images or geojson files, query-able info we want to store on-chain, like we have [here](https://github.com/regen-network/regen-registry-ops/blob/cb68947a7b4416776c0fc9a8289a709f6ab53d07/C01/project-metadata/C01-ProjectId934-metadata.json). With that in mind, we might want to take out any non-query-able info from the `C01-verified-carbon-standard-project.ttl`, so anything in the `regen:ProjectPageMediaGroup`. 

I know this adds on some extra work (possibly outside this PR), but I think it might be best in the long run. Thoughts?

shacl/projects/C01-verified-carbon-standard-project.ttl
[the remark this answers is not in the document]
blushi: yes for the project developer image for specifically (that can be individual or org)

shacl/credit-batches/C01-verified-carbon-standard-batch.ttl
```diff
@@ -26,3 +26,18 @@ regen:CreditBatchShape a sh:NodeShape ;
     sh:maxCount 1 ;
   ] ;
 .
+
+regen:NameOptionalUrlShape a sh:NodeShape ;
```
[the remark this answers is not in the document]
blushi: For credit batches, as it is specified currently in this SHACL schema, you need at least one additional certification but the URL for the certification is optional (you can only provide a name), which seems to be aligned with what you say

Then whether adding at least one additional certification is mandatory or not, this is just what was in the original SHACL graph that Kyle had worked on some months ago. But looking now at the figma: https://www.figma.com/file/TyhFaXqko9F8OvLEE4Zomg/Issuer-Create%2FEdit-Projects%2C-Credit-Class%2C-Batches?node-id=74%3A24702, I see that Erika made the whole field optional, thoughts?

blushi:
> Hey @blushi @wgwz, thanks for the patience on my responses as I get back from the conference. I spent some time with @clevinson thinking through how we're structuring metadata for project pages and here are our thoughts.
> 
> Generally speaking, the only information we should be storing in our on-chain metadata IRIs are information and relationships we want to query against. Right now the information and outlined in the credit class and credit batch seems pretty straight forward as most users will want to know this like, which methodology is used by a credit class or what the retirement serial number is for a certain batch. For projects it should be similar, and we should only store info want to query against, like the size or location, methodology used, start & end date, etc...
> 
> With that in mind, things like images and videos should probably not be included in the metadata IRIs we store on chain. Obviously this adds in some complexity in the way we are currently structuring our SHACL schemas and building our project pages as we've blended much of the info we store off-chain, such as images or geojson files, query-able info we want to store on-chain, like we have [here](https://github.com/regen-network/regen-registry-ops/blob/cb68947a7b4416776c0fc9a8289a709f6ab53d07/C01/project-metadata/C01-ProjectId934-metadata.json). With that in mind, we might want to take out any non-query-able info from the `C01-verified-carbon-standard-project.ttl`, so anything in the `regen:ProjectPageMediaGroup`.
> 
> I know this adds on some extra work (possibly outside this PR), but I think it might be best in the long run. Thoughts?

So just to be sure, we only want to pull out from the metadata referenced on-chain everything related to Media, right?
What about the project developer profile image? 

Right now, all the metadata is stored in the `project` table `metadata` column anyway and not referenced on-chain, so I don't believe we need to change the way things work currently.
But later on, when we have on-chain projects, we can store only "query-able" info in the `metadata_graph` table (just as we do currently for on-chain credit class/batch metadata) and keep storing the rest of the info related to media within the off-chain `project` table. This means we'll need to adapt:
- this project SHACL graph: in particular, I believe we could have one SHACL shape for C01 query-able project metadata and another one for just media data
- the way we query project data from the project page (we'll need to do this anyway once we have on-chain projects)
- the project creation form logic to save the query-able metadata in `metadata_graph` table and the rest in `project.metadata`

shacl/projects/C01-verified-carbon-standard-project.ttl
blushi: Ok updated as part of https://github.com/regen-network/regen-registry-standards/pull/16/commits/4f7e38808bb918fa69605509127e2b2c59e88106 (also did the same for C01 credit batch)
this means we should indeed update the project metadata in the ops repo but also the metadata we store in the db to be consistent, should we create a separate issue to do so?

blushi: > 



> > but the thing is that we've already issued some credit batches on-chain using regen:CreditBatch as @type and those can't be updated...
> 
> could you remind me why those can't be updated?
> 
> but that's unfortunate. maybe we just have to accept that `Project` means `C01-Project`. And maybe in the future, i.e. `regen:C02-Project` is a subclass of `regen:BaseProject` (`Project` could also be a subclass of `regen:BaseProject`)

we can edit the project metadata so I don't think that's the issue, I believe it would be confusing to just assume `regen:Project` = `regen:C01-Project` given that.
the only issue is with regards to existing credit batches metadata that we can't update on-chain but as @S4mmyb said, probably that's fine to keep those existing ones with `regen:CreditBatch` type and then have future C01 batches use `regen:C01-CreditBatch` as long as that's documented

shacl/credit-batches/C01-verified-carbon-standard-batch.ttl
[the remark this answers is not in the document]
S4mmyb: Ooh cool, thanks for the explanation! And yeah I think we do want the entire field optional

shacl/projects/C01-verified-carbon-standard-project.ttl
S4mmyb: > Ok updated as part of [4f7e388](https://github.com/regen-network/regen-registry-standards/commit/4f7e38808bb918fa69605509127e2b2c59e88106) (also did the same for C01 credit batch) this means we should indeed update the project metadata in the ops repo but also the metadata we store in the db to be consistent, should we create a separate issue to do so?

Yeah I think a separate issue for this makes the most sense as to not spend too much time falling down the rabbit hole on this update.

S4mmyb:
> So just to be sure, we only want to pull out from the metadata referenced on-chain everything related to Media, right? What about the project developer profile image?
> 
> Right now, all the metadata is stored in the `project` table `metadata` column anyway and not referenced on-chain, so I don't believe we need to change the way things work currently. But later on, when we have on-chain projects, we can store only "query-able" info in the `metadata_graph` table (just as we do currently for on-chain credit class/batch metadata) and keep storing the rest of the info related to media within the off-chain `project` table. This means we'll need to adapt:
> 
> * this project SHACL graph: in particular, I believe we could have one SHACL shape for C01 query-able project metadata and another one for just media data
> * the way we query project data from the project page (we'll need to do this anyway once we have on-chain projects)
> * the project creation form logic to save the query-able metadata in `metadata_graph` table and the rest in `project.metadata`

Hmm, that's a good question. I think maybe not as it's not querable and we want to stay consistent, but it would be much cleaner just to include it the way you have it so I'm split on this one. What do you think @blushi @wgwz ?

On a somewhat related note, during our retreat we talked about introducing DiDs associated with regen addresses where you could "verify" the organization or entity claiming claims against certain actions. There were some ideas circulating about anchoring a a website, twitter, or linkedin profile, to an address to provide some sort of verification of the individual/org on the other end and I'm wondering if where we would put the image in that scenario.

shacl/credit-batches/C01-verified-carbon-standard-batch.ttl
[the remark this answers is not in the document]
blushi: ok made the entire field optional as part of https://github.com/regen-network/regen-registry-standards/pull/16/commits/d915b681adda69f0c3374dfc8d41b0f420d48819

shacl/projects/C01-verified-carbon-standard-project.ttl
blushi: added: https://github.com/regen-network/regen-registry/issues/941

wgwz:
> Hmm, that's a good question. I think maybe not as it's not querable and we want to stay consistent, but it would be much cleaner just to include it the way you have it so I'm split on this one. What do you think @blushi @wgwz ?

@S4mmyb based on what you raised in comments and how @blushi is thinking about that, I think that it makes sense to me. As I was reading Marie's comments, I was thinking that it could be helpful to have on-chain and off-chain versions of these schemas. I just wanted to propose that as a way to phrase the difference.

Also sorry for falling off on the whole C01-project discussion. I'm glad to see we've opened another issue for that. I think I'm missing some understanding on that thread but I guess we can discuss further as we handle this new task (https://github.com/regen-network/regen-registry/issues/941)

wgwz (approved):

S4mmyb (approved):

wgwz:
@S4mmyb sounds good! i'll be sure to add a topic for this.

PR #24, opened 2022-08-25
Kyle/1121 cfc metadata

wgwz:
## Description

Closes: https://github.com/regen-network/regen-registry/issues/1121

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/ops/C01/credit-class-metadata/C01-verified-carbon-standard-credit-class.jsonld
+++ b/ops/C01/credit-class-metadata/C01-verified-carbon-standard-credit-class.jsonld
@@ -1,41 +1,33 @@
 {
-  "@context":
-  {
+  "@context": {
     "schema": "http://schema.org/",
     "regen": "http://regen.network/",
     "xsd": "http://www.w3.org/2001/XMLSchema#",
-    "schema:itemListElement":
-    {
+    "schema:itemListElement": {
       "@container": "@list"
     },
-    "regen:sectoralScope":
-    {
+    "regen:sectoralScope": {
       "@container": "@list"
     },
-    "regen:offsetGenerationMethod":
-    {
+    "regen:offsetGenerationMethod": {
       "@container": "@list"
     }
   },
   "@type": "regen:CreditClass",
   "schema:name": "Verified Carbon Standard",
   "schema:description": "This credit class provides a vehicle for nature based Verified Carbon Units (VCUs) to enter the blockchain space via issuance on Regen Ledger. It can be used by project developers, credit brokers, and other stakeholders interested in digitizing carbon credits issued by the VCS program to make them available in the emerging world of decentralized finance.",
-  "schema:url":
-  {
+  "schema:url": {
     "@type": "schema:URL",
     "@value": "https://library.regen.network/v/regen-registry-credit-classes/credits-from-other-registries/verified-carbon-standard-credit-class"
   },
-  "regen:sourceRegistry":
-  {
+  "regen:sourceRegistry": {
     "schema:name": "Verra",
-    "schema:url":
-    {
+    "schema:url": {
       "@type": "schema:URL",
       "@value": "https://verra.org/"
     }
   },
-  "regen:sectoralScope":
-  [
+  "regen:sectoralScope": [
     {
       "@type": "xsd:string",
       "@value": "Agriculture Forestry and Other Land Use (AFOLU)"
@@ -45,8 +37,7 @@
       "@value": "Livestock and Manure Management"
     }
   ],
-  "regen:offsetGenerationMethod":
-  [
+  "regen:offsetGenerationMethod": [
     {
       "@type": "xsd:string",
       "@value": "Emission Reductions"
@@ -56,11 +47,9 @@
       "@value": "Carbon Removals"
     }
   ],
-  "regen:approvedMethodologies":
-  {
+  "regen:approvedMethodologies": {
     "@type": "schema:BreadcrumbList",
-    "schema:itemListElement":
-    [
+    "schema:itemListElement": [
       {
         "@type": "regen:Methodology",
         "schema:name": "Afforestation and reforestation of degraded land",
@@ -814,8 +803,7 @@
         }
       }
     ],
-    "schema:url":
-    {
+    "schema:url": {
       "@type": "schema:URL",
       "@value": "https://library.regen.network/v/regen-registry-credit-classes/credits-from-other-registries/verified-carbon-standard-credit-class/approved-methodologies"
     }
--- a/ops/C01/project-metadata/C01-ProjectId612-metadata.jsonld
+++ b/ops/C01/project-metadata/C01-ProjectId612-metadata.jsonld
@@ -29,6 +29,15 @@
     }
   },
   "regen:offsetGenerationMethod": "Emission Reduction",
+  "regen:renderSpecification": [{
+    "key": "regen:offsetGenerationMethod",
+    "label": "Offset Generation Method",
+    "index": 0
+  }, {
+    "key": "regen:projectActivity",
+    "label": "Project Activity",
+    "index": 1
+  }],
   "regen:vcsMethodology": {
     "schema:name": "Methodology for Avoided Ecosystem Conversion",
     "schema:url": {
--- a/ops/C02/credit-batch-metadata/C02-001-2019-batch.jsonld
+++ b/ops/C02/credit-batch-metadata/C02-001-2019-batch.jsonld
@@ -1,0 +1,10 @@
+{
+  "@context": {
+    "schema": "http://schema.org/",
+    "regen": "http://regen.network/"
+  },
+  "@type": "regen:C02-CreditBatch",
+  "regen:creditSerialNumbers": ["6477 - 6576", "6677 - 6841"],
+  "regen:verificationReport": "https://www.notion.so/regennetwork/King-County-Urban-Forest-Preservation-Program-ce67afcfbd9b4daa8240186ddc160454",
+  "regen:vintageYear": "2019"
+}
--- a/ops/C02/credit-batch-metadata/C02-002-2020-batch.jsonld
+++ b/ops/C02/credit-batch-metadata/C02-002-2020-batch.jsonld
@@ -1,0 +1,10 @@
+{
+  "@context": {
+    "schema": "http://schema.org/",
+    "regen": "http://regen.network/"
+  },
+  "@type": "regen:C02-CreditBatch",
+  "regen:creditSerialNumbers": ["20465 - 21964"],
+  "regen:verificationReport": "https://www.cityforestcredits.org/wp-content/uploads/2019/10/Evergreen-Verification-Report.pdf",
+  "regen:vintageYear": "2020"
+}
--- a/ops/C02/credit-batch-metadata/C02-005-2022-batch.jsonld
+++ b/ops/C02/credit-batch-metadata/C02-005-2022-batch.jsonld
@@ -1,0 +1,10 @@
+{
+  "@context": {
+    "schema": "http://schema.org/",
+    "regen": "http://regen.network/"
+  },
+  "@type": "regen:C02-CreditBatch",
+  "regen:creditSerialNumbers": ["78240 - 84648"],
+  "regen:verificationReport": "https://www.cityforestcredits.org/wp-content/uploads/2021/10/Harvey-Manning-Verification-Report-Final.pdf",
+  "regen:vintageYear": "2022"
+}
--- a/ops/C02/credit-batch-metadata/C02-003-2021-batch.jsonld
+++ b/ops/C02/credit-batch-metadata/C02-003-2021-batch.jsonld
@@ -1,0 +1,10 @@
+{
+  "@context": {
+    "schema": "http://schema.org/",
+    "regen": "http://regen.network/"
+  },
+  "@type": "regen:C02-CreditBatch",
+  "regen:creditSerialNumbers": ["32524 - 41238"],
+  "regen:verificationReport": "https://www.notion.so/regennetwork/St-Elmo-Preservation-Project-9498fc21d024404190f3f0b24a3351b0",
+  "regen:vintageYear": "2021"
+}
--- a/ops/C02/credit-batch-metadata/C02-004-2021-batch.jsonld
+++ b/ops/C02/credit-batch-metadata/C02-004-2021-batch.jsonld
@@ -1,0 +1,10 @@
+{
+  "@context": {
+    "schema": "http://schema.org/",
+    "regen": "http://regen.network/"
+  },
+  "@type": "regen:C02-CreditBatch",
+  "regen:creditSerialNumbers": ["43432 - 47570"],
+  "regen:verificationReport": "https://www.cityforestcredits.org/wp-content/uploads/2020/10/Bainbridge-Forest-Verification-Report.pdf",
+  "regen:vintageYear": "2021"
+}
--- a/ops/C02/credit-batch-metadata/C02-006-2022-batch.jsonld
+++ b/ops/C02/credit-batch-metadata/C02-006-2022-batch.jsonld
@@ -1,0 +1,10 @@
+{
+  "@context": {
+    "schema": "http://schema.org/",
+    "regen": "http://regen.network/"
+  },
+  "@type": "regen:C02-CreditBatch",
+  "regen:creditSerialNumbers": ["7545 - 9444", "12256 - 12429"],
+  "regen:verificationReport": "https://www.cityforestcredits.org/wp-content/uploads/2020/02/Buena-Vista-Heights-Project-Verification-Report.pdf",
+  "regen:vintageYear": "2022"
+}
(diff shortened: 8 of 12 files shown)


shacl/credit-batches/C01-verified-carbon-standard-batch.ttl
```diff
@@ -6,8 +6,6 @@
 @prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
 @prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
 
-regen:regen:C01-CreditBatch rdfs:subClassOf regen:CreditBatch .
```
wgwz, line 9: this line was incorrect and as far as i could tell it wasn't actually having any effect on shacl validation.

shacl/credit-batches/credit-vintage.ttl
```diff
@@ -1,3 +1,4 @@
+@prefix dash: <http://datashapes.org/dash#> .
```
wgwz, line 1: adds a missing prefix.

shacl/projects/project-page.ttl
```diff
@@ -112,6 +112,7 @@ regen:ProjectPageShape
   ) ;
   # We should show at least one of the project stakeholders on the project page.
   sh:property [
+    sh:path regen:projectStakeholder ;
```
wgwz, line 115: adds a field that was missing.

shacl/projects/C01-verified-carbon-standard-project.ttl
```diff
@@ -37,7 +37,7 @@ regen:C01-ProjectShape a sh:NodeShape ;
   ] ;
   sh:property [
     sh:path regen:projectDeveloper ;
-    sh:node regen:RoleShape ;
+    sh:node regen:ProjectRoleShape ;
```
wgwz, line 40: renamed because this was in conflict with another shape in a different SHACL schema, and they are not the same.

ops/C01/credit-class-metadata/C01-verified-carbon-standard-credit-class.jsonld
```diff
@@ -1,41 +1,33 @@
...
       "@container": "@list"
     },
-    "regen:offsetGenerationMethod":
-    {
+    "regen:offsetGenerationMethod": {
       "@container": "@list"
     }
   },
-  "@type": "regen:CreditClass",
+  "@type": "regen:C01-CreditClass",
```
wgwz, line 16: important: this change will require updating the on-chain C01 credit class metadata.

wgwz: the other diffs in this file are just formatting changes.

ops/C01/project-metadata/C01-ProjectId612-metadata.jsonld
```diff
@@ -17,6 +17,10 @@
   "regen:projectDeveloper": {
     "@type": "regen:OrganizationDisplay",
     "schema:name": "Wildlife Works Carbon LLC",
+    "schema:logo": {
```
wgwz, line 20: important: this field was specified in our C01 shacl schema, and i added it here. if we have a value for this we can add it. otherwise, if we can remove this field from the shacl schema if it's no longer required.

ops/C01/project-metadata/C01-ProjectId612-metadata.jsonld
```diff
@@ -29,6 +33,15 @@
     }
   },
   "regen:offsetGenerationMethod": "Emission Reduction",
+  "regen:renderSpecification": [{
```
wgwz, line 36: note-to-self: remove.

wgwz:
this is just a note that we might be able to use this method for sort of "inheriting" certain properties of other shapes:

https://book.validatingrdf.com/bookHtml011.html#sec126

/cc @mhagel

ops/C02/credit-batch-metadata/C02-002-2020-batch.jsonld
```diff
@@ -0,0 +1,16 @@
+{
+  "@context": {
+    "schema": "http://schema.org/",
+    "regen": "http://regen.network/",
+    "xsd": "http://www.w3.org/2001/XMLSchema#",
+    "regen:cfcVintageYear": {"@type": "xsd:gYear"}
+  },
+  "@type": "regen:C02-CreditBatch",
+  "regen:cfcCreditSerialNumbers": ["20465 - 21964"],
```
clevinson, line 9: Maybe I'm missing context, but just curious -- should this be unpacked into a list of integers?

clevinson: Or maybe richer custom "range" object of sorts?

ops/C02/credit-batch-metadata/C02-005-2022-batch.jsonld
```diff
@@ -0,0 +1,16 @@
+{
+  "@context": {
+    "schema": "http://schema.org/",
+    "regen": "http://regen.network/",
+    "xsd": "http://www.w3.org/2001/XMLSchema#",
+    "regen:cfcVintageYear": {"@type": "xsd:gYear"}
```
clevinson, line 6: Could we pull out this definition into its own file (hosted on Regen.network) so that we don't have to include in each project's jsonld?

ops/C02/credit-batch-metadata/C02-002-2020-batch.jsonld
wgwz: this is how it was listed in the document i was copying the metadata from, we’ll need some feedback from @S4mmyb on this. i took it as this string is literally there serial number format, rather than a range or something like that.

ops/C02/credit-batch-metadata/C02-005-2022-batch.jsonld
wgwz: regardless of that i think we would need to have the LOC that you commented on for this to be considered valid RDF (unless i’m missing something). but yes, we should talk about hosting our ontology/schema public-ally somewhere soon. apache jena has a tool for doing that, which i think could work nicely with our current shacl schema.

ops/C01/project-metadata/C01-ProjectId612-metadata.jsonld
mhagel: I think logo should be optional. min: 0 instead of min:1?

jsonld/projects/C02-project.jsonld
```diff
@@ -0,0 +1,75 @@
...
+  "regen:jurisdiction": "",
+  "schema:description": "",
+  "regen:projectOperator": {
+    "schema:name": "Allegheny Land Trust",
+    "schema:url": ""
+  },
+  "regen:projectType": "",
+  "regen:projectActivity": "",
+  "regen:offsetGenerationMethod": "",
+  "regen:projectSize": {
```
S4mmyb, line 47: should be filled in by the interface

jsonld/projects/C02-project.jsonld
```diff
@@ -0,0 +1,75 @@
...
+    "qudt:unit": "",
+    "qudt:numericValue": 0
+  },
+  "regen:projectStartDate": "",
```
S4mmyb, line 51:
```suggestion
  "regen:projectStartDate": {
     "@type": "xsd:date",
     "@value": ""
  },
```

let's try and use `xsd:date` here

jsonld/projects/C02-project.jsonld
```diff
@@ -0,0 +1,75 @@
...
+  "regen:projectEndDate": "",
```
S4mmyb, line 52:
```suggestion
  "regen:projectEndDate": {
     "@type": "xsd:date",
     "@value": ""
  },
```

^^

jsonld/projects/C02-project.jsonld
```diff
@@ -0,0 +1,75 @@
...
+    "place_name": "",
+    "geometry": {
+      "type": "Point",
+      "coordinates": [
+        0,
+        0
+      ]
+    }
+  },
+  "regen:offsetProtocol": ""
```
S4mmyb, line 74: I think we actually want this to be a link so maybe: 
```suggestion
  "regen:offsetProtocol": {
    "schema:name": "",
    "schema:url": {
      "@type": "schema:URL",
      "@value": ""
    }
  }
```

S4mmyb: the "offset protocol" is the same as methodology so we want to track name and url (idk if version is important)

ops/C02/credit-batch-metadata/C02-001-2019-batch.jsonld
```diff
@@ -3,14 +3,29 @@
...
+    "regen:verificationReports": {
+      "@container": "@list"
+    },
+    "schema:url": {
+      "@type": "schema:URL"
+    }
   },
   "@type": "regen:C02-CreditBatch",
-  "regen:cfcCreditSerialNumbers": ["6477 - 6576", "6677 - 6841"],
+  "regen:cfcCreditSerialNumbers": [
```
S4mmyb, line 20: @wgwz do you know if there's a cleaner way to do this? I think it would be a bit cleaner to not represent serial numbers as a list of strings?

ops/C02/credit-batch-metadata/C02-002-2020-batch.jsonld
```diff
@@ -3,14 +3,28 @@
...
+  "regen:cfcCreditSerialNumbers": [
+    "20465 - 21964"
+  ],
   "regen:cfcVintageYear": 2020,
-  "regen:cfcProjectId": "",
-  "regen:verificationReports": [{
-    "@type": "schema:URL",
-    "@value": "https://www.cityforestcredits.org/wp-content/uploads/2019/10/Evergreen-Verification-Report.pdf"
-  }]
+  "regen:cfcProjectId": "TODO",
```
S4mmyb, line 24:
```suggestion
```

We probably don't need to include this as the project will be listed on the batch-info page and they could link out to that, right?

S4mmyb (changes requested):

S4mmyb:
Hey @wgwz , thanks for the PR! I requested a few changes, mostly in the templated docs. For the most part it looks good - I think one thing we want to make sure of is that when the the user flow to create a project page is being used it correctly fills out the fields in the json-ld being populated behind the scenes. This includes: 

- `projectName`
- `projectDecription`
- `projectSize`
- `projectOperator` (via the roles section)
- `projectLocation`

I also noticed that we don't have in here a section for media or project photos. I'm sure this is being automatically populated in the user-interface, we just need to make sure to include it in the schema validation in the same way we do for C01 project pages. 

Overall great job!

jsonld/projects/C02-project.jsonld
```diff
@@ -0,0 +1,75 @@
...
+    },
+    "regen:projectStartDate": {
+      "@type": "xsd:date"
+    },
+    "regen:projectEndDate": {
+      "@type": "xsd:date"
+    }
+  },
+  "@type": "regen:C02-Project",
+  "schema:name": "",
```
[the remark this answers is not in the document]
wgwz: > I think we should get in the habit of wrapping this sort of stuff in descriptors, in this case projectName, so that users can understand what we're referring to as schema:name might be ambiguous for those who didn't write the schema

@S4mmyb I think that this change might not be necessary, and here is why I'm thinking that. An important point of making use of schema.org's definition of "name" is that it represents something general enough that it can be useful across different things. For example, the "name" property of a schema.org "User" or "Organization" use the same property ("name"). They do not have a "user name" or an "org name". They have a statement saying "this user has a name". 

While I do want to make these schema's as user-friendly as possible, unfortunately at this level, I think we are stuck with the complexities of RDF data (for good reasons as we know). Here's something I was looking into which is tangential but related in the sense that, RDF is a hurdle. I.e. I found it helpful taking a look at the documentation for [google's SEO stuff related to json-ld](https://developers.google.com/search/docs/advanced/structured-data/intro-structured-data). 

 I think the point I'm trying to make is that there is a necessary learning curve to writing JSON-LD. And even a huge organization using RDF like google, where it provides some pretty important features for users, basically sends you down a learning path, and to learn by example.
 
I think that may be the path of least resistance for us as well. We are building out a catalogue of examples, and we are trying to do our best to develop a standard, while at the same time providing a flexible data model. I think showcasing these will probably be most helpful for the "user" interacting with the metadata at this level. Just like adding JSON-LD to your blog requires some technical skill and digging, so to will this IMO. 

> This should be something filled by the user interface, correct?

No, not for the initial C02 projects but in the future yes. @mhagel and I have agreed that we will use the basic project creation for C02, where we will just paste the manually created metadata into a text box and submit it. We agreed to try to make the C01/C02 project metadata as similar as possible, so that modifying the full project creation flow will be easier.

We decided to take this path because it felt like a tricky change prior to marketplace go-live. We basically wanted/needed some more time to think through how we would handle the few differences in the C02 metadata in the create project flow.

jsonld/projects/C02-project.jsonld
wgwz: we don't need the `@type` here because it's been included in the `@context` section.

jsonld/projects/C02-project.jsonld
wgwz: we don't need the `@type` here because it's been included in the `@context` section.

ops/C02/credit-batch-metadata/C02-002-2020-batch.jsonld
wgwz: i just put it here because it was in the notion doc as a field for the credit batches metadata. glad to remove if it's not needed

wgwz:
> I also noticed that we don't have in here a section for media or project photos. I'm sure this is being automatically populated in the user-interface, we just need to make sure to include it in the schema validation in the same way we do for C01 project pages.

@S4mmyb we have `previewPhoto` and `galleryPhotos` fields (same fields as C01) for both of the two projects i've written the metadata for so far.

```
  "regen:previewPhoto": "https://regen-registry.s3.amazonaws.com/projects/C02/harvey-manning-01.jpg",
  "regen:galleryPhotos": [
    "https://regen-registry.s3.amazonaws.com/projects/C02/harvey-manning-02.jpg",
    "https://regen-registry.s3.amazonaws.com/projects/C02/harvey-manning-03.jpg"
  ],
```

ops/C02/credit-batch-metadata/C02-001-2019-batch.jsonld
wgwz: yeah, Cory had a similar comment above, and honestly it's just something i've been meaning to ask you about. so i guess this means that those number ranges represent the full list of possible "credit serial numbers". 

```
6477
6478
6479
...
...
6576
```

```
6677
6678
6679
...
...
6841
```

the simple solution takes up a bunch of space, write out the full list of all these serial numbers as integers. i don't know of a standard way of representing a range of values in a string. the method we're using seems pretty reasonable, i could add a better validation of that particular format.

this is a totally different context but here is [an ISO-standard](https://en.wikipedia.org/wiki/ISO_8601#Time_intervals) for representing an interval of time. TL;DR `<start>/<end>` (which is basically just a few characters off from we have.

jsonld/projects/C02-project.jsonld
wgwz: could we perhaps take it a step further and unify "regen:offsetGenerationMethod" and "regen:offsetProtocol" so that we de-duplicate this field between C01 and C02 project metadata?

the question is are these fundamentally different things?

wgwz: change `regen:offsetProtocol` and `regen:methodology`, and this so means we're going add a `schema:version` and `schema:id` (CFC doesn't have version or ID though) but these will be optional fields.

wgwz:
todo: remove the `regen:cfcProjectId`

ops/C02/credit-batch-metadata/C02-002-2020-batch.jsonld
wgwz: @S4mmyb and I talked about this yesterday, and we agree it's OK to keep the convention of "<start> - <end>". AFAICT we have two choices for serialization; a list of integers or a some string representation of the range of values. e.g. ISO-8601 represents time ranges in strings as "<start-dt>/<end-dt>".

S4mmyb:
@wgwz I just gave this a final review and it all looks good! Changes/updates I made: 

- Small tweaks on content (project location lat/lon, descriptions, etc...) to better reflect what CFC has on their project pages + adding dates/offset protocols to the projects which didn't have them
- Changed batch file names for the batches we plan on issuing first to reflect the format seen on the ledger: `<creditClassID-projectID-batchStartDate-batchEndDate-batchID`. Note I used the start and end dates in the verification report as the batch start/end dates. 
- Added project verifier to the C02 batches. I updated the shacl shema, json, and ops files to reflect this using the `NameURLShape` and ran the validation script and everything conformed (yahoo!)

I'll go ahead and approve this, but before we merge we just need to make sure to add in the project start/end date to the King County project (`C01-001`) which is missing from the project design document. I emailed the CFC team to get those dates - once we have them all we need to do is fill them in and make sure it successfully conforms to the shape as it's currently failing on invalid date format (ie we don't have a date). 

Once again, great work!

S4mmyb:
Added in the final dates for the final C02 project `C02-001` and all checks on the `shacl-validate` have passed so approving now cc @wgwz

S4mmyb (approved):

Issue #20, opened 2022-08-11
Use versioning and tag versions within repository


ryanchristo:
## Summary

It would be nice if each schema had a regen registry standard version specified within the schema and versions were tagged within the repository so that specific version could be referenced.

This may require retroactively tagging a version initially used for classes and batches (a version that does not include the regen registry standard version within the schemas and which was previously used for metadata on Regen Ledger).

This would also require finalizing the class, project, and batch schemas that will be used with any new classes, projects, and batches (including adding a regen registry standard version), and then tagging a new version.

Using versions will allow projects to choose a specific version and migrate to a new version when a new version is finalized. Currently we are using the latest changes and would like to commit to using an explicit version knowing we may need to update to a new version in the future and that we would make changes when a new version has been finalized.

ryanchristo:
cc @clevinson @S4mmyb

wgwz:
@ryanchristo i think this makes sense, because then we would know which schema we expect to validate successfully against a given piece of metadata. given a piece of remote metadata you know which schema must be applied to it by this version number. and you have the ability to find the schema corresponding to this version number.

so we would add a field named something like `regen:standardsVersion` to the batch, class and project schema. i guess we would just have to figure out precisely what type of versioning scheme to use, and figure out how we want to track it. i.e. would the `regen:standardsVersion` be something that we can increment independently for credit, batch and project? (i think the answer is probably no but it's a detail to sort through). and would we use a git tag or just add versions to filenames?

i.e. using `regen:standardsVersion` field and adding versions to filenames with the assumption that `regen:standardsVersion` must be incremented uniformly:

```
{
  "@context": {
    "schema": "http://schema.org/",
    "regen": "http://regen.network/",
    "regen:additionalCertifications": { "@container": "@list" }
  },
  "@type": "regen:C01-CreditBatch",
  "regen:standardsVersion": "v0.0.1",
  "regen:vcsRetirementSerialNumber": "",
  "regen:vcsProjectId": "",
  "regen:additionalCertifications": [
    {
      "schema:name": "",
      "schema:url": {
        "@type": "schema:URL",
        "@value": ""
      }
    },
    {
      "schema:name": "",
      "schema:url": {
        "@type": "schema:URL",
        "@value": ""
      }
    }
  ]
}
```

```
❯ ls -l shacl/credit-batches 
total 16
C01-verified-carbon-standard-batch.v0.0.1.ttl
[...]

PR #32, opened 2022-10-27
Add toucan metadata for project and credit batch

wgwz:
## Description

These additions are based on [this spec](https://www.notion.so/regennetwork/Finalized-Mapping-for-Toucan-Metadata-including-operational-additions-4f0574ad35924f6e9c700a732450eba2).

The most important files to review are:

- [`ops/C03/project-metadata/C03-project-minimal-example.jsonld`](https://github.com/regen-network/regen-registry-standards/blob/kyle/add-toucan-metadata/ops/C03/project-metadata/C03-project-minimal-example.jsonld)
  - the bare minimum that we'd expect to receive for a toucan created project
- [`ops/C03/project-metadata/C03-project-full-example.jsonld`](https://github.com/regen-network/regen-registry-standards/blob/kyle/add-toucan-metadata/ops/C03/project-metadata/C03-project-full-example.jsonld)
  - an example of a fully-complete touch project metadata
- [`ops/C03/credit-batch-metadata/C03-batch-minimal-example.jsonld`](https://github.com/regen-network/regen-registry-standards/blob/kyle/add-toucan-metadata/ops/C03/credit-batch-metadata/C03-batch-minimal-example.jsonld)
  - what we expect the touch credit batch metadata to be. (at present we only have a minimal version of this)

If you know of more realistic data values for any of the examples above, please comment on those and let me know what they should be.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


jsonld/projects/C03-toucan-verified-carbon-standard-project.jsonld
```diff
@@ -0,0 +1,46 @@
...
+    },
+    "regen:projectStartDate": {
+      "@type": "xsd:date"
+    },
+    "regen:projectEndDate": {
+      "@type": "xsd:date"
+    }
+  },
+  "@type": "regen:C03-Project",
+  "regen:toucanProjectTokenId": "",
```
clevinson, line 26:
```suggestion
  "regen:toucanProjectTokenId": 0,
```
I believe this will be a number

jsonld/projects/C03-toucan-verified-carbon-standard-project.jsonld
```diff
@@ -0,0 +1,46 @@
...
+  "schema:name": "",
+  "regen:vcsProjectId": 0,
```
clevinson, line 28:
```suggestion
  "regen:vcsProjectId": "",
```

I think this will be a string (e.g. `"VCS-123"`)

ops/C03/project-metadata/C03-project-full-example.jsonld
```diff
@@ -0,0 +1,46 @@
...
+    },
+    "regen:projectStartDate": {
+      "@type": "xsd:date"
+    },
+    "regen:projectEndDate": {
+      "@type": "xsd:date"
+    }
+  },
+  "@type": "regen:C03-Project",
+  "regen:toucanProjectTokenId": "123",
```
clevinson, line 26:
```suggestion
  "regen:toucanProjectTokenId": 123,
```

I think this can be a number

ops/C03/project-metadata/C03-project-full-example.jsonld
```diff
@@ -0,0 +1,46 @@
...
+  "schema:name": "VCS-123",
+  "regen:vcsProjectId": 123,
```
clevinson, line 28:
```suggestion
  "regen:vcsProjectId": "VCS-123",
```

From Toucan we get a string here, so I think its probably easiest to pass it along with the VCS prefix.   Then again, I'm also wanting to balance being easily compatible with Toucan and using the same data types as in `C01`...  There's trade-offs in both directions I think.  I would hate for us to get a parsing error on the Toucan side and not be able to set this though if our regex lookup is looking for `VCS-XXX` and fails bc it changes to `VCS:XXX` at some point.

shacl/projects/C03-project.ttl
```diff
@@ -0,0 +1,59 @@
...
+  sh:property regen:ProjectOptionalStartDatePropertyShape ;
+  sh:property regen:ProjectOptionalEndDatePropertyShape ;
+.
+
+regen:ToucanProjectTokenIDPropertyShape sh:path regen:toucanProjectTokenId ;
+  sh:datatype xsd:string ;
+  sh:minCount 1 ;
+  sh:maxCount 1 ;
+  sh:minLength 1;
+.
```
clevinson, line 59: I think this can be an `xsd:interger`

clevinson (commented):
Thanks @wgwz ! Got a few questions / comments in-line, but generally this is looking good.  Excited to be having @ryanchristo take a look at it to so we get some more ledger <> interfaces collaboration on the metadata schema stuff :)

ops/C03/project-metadata/C03-project-minimal-example.jsonld
```diff
@@ -0,0 +1,37 @@
...
+    },
+    "schema:version": {
+      "@type": "schema:Text"
+    },
+    "regen:projectStartDate": {
+      "@type": "xsd:date"
+    },
+    "regen:projectEndDate": {
+      "@type": "xsd:date"
+    }
```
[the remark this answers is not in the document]
clevinson: This question is specifically w/r/t the `minimal-example.jsonld`

ops/C03/project-metadata/C03-project-full-example.jsonld
```diff
@@ -0,0 +1,46 @@
...
+  "@context": {
+    "schema": "http://schema.org/",
+    "regen": "http://regen.network/",
+    "xsd": "http://www.w3.org/2001/XMLSchema#",
+    "regen:approvedMethodologies": {
+      "@container": "@list"
+    },
+    "schema:identifier": {
+      "@type": "schema:Text"
+    },
```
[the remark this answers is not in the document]
wgwz: > Maybe we should spin this out into our own issue and treat as a separate discussion? what is the best way to proceed from your perspective?

@clevinson i think the best way to proceed is with the `xsd:string` type. it results in a simpler JSON structure, and it is still valid RDF as evidenced by our shacl schema and the validations we are running there. i don't think we should be overly concerned about strict conformance to the schema.org ontology/vocabulary down to this level of granularity. by the metric of simpler being easy to maintain and make use of, i think it's forward looking. 🤷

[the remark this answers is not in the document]
wgwz: > I'd love to figure out how to verify for myself (with apache jena or using our iri-gen package) that these are actually necessary, and that string values aren't implicitly interoperable / equivalent with explicit schema:Text types. can you point to what docs you've read that give you this understanding? Or how you've done manual checks of this in the past?

@clevinson yeah, i will take down todo item to make a screenshare that illustrates how to see this with apache jena by modifying the shacl/json-ld and running the validations. i'd love feedback on the verification setup i've put in place, tried to keep it as simple as possible so far.

shacl/methodology.ttl
```diff
@@ -15,22 +15,24 @@ regen:MethodologyVersionShape
...
+    sh:maxCount 1 ;
     sh:minLength 1 ;
   ] ;
   sh:property [ 
     sh:path schema:identifier ;
     sh:datatype schema:Text ;
-    sh:minCount 1;
-    sh:maxCount 1;
+    sh:minCount 1 ;
+    sh:maxCount 1 ;
```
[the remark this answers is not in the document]
wgwz: with the changes added here:

- `identifier` is still required
- `name` is now required
- `url` is still optional
- `version` is now optional

this is backwards compat with our current metadata (as evidenced by passing shacl validations)

ops/C03/project-metadata/C03-project-minimal-example.jsonld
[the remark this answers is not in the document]
wgwz: no, we don't have to include these. i was on the fence about keeping them. but now that you point it out i think it's a good idea to remove them.

shacl/projects/C03-project.ttl
```diff
@@ -0,0 +1,59 @@
...
+  sh:property regen:ProjectActivityOptionalPropertyShape ;
+  sh:property regen:ProjectTypeOptionalPropertyShape ;
+  sh:property [
+    sh:path regen:offsetGenerationMethod ;
+    sh:maxCount 1 ;
+    sh:minLength 1 ;
+    sh:datatype xsd:string ;
+  ] ;
+  sh:property regen:ProjectOptionalStartDatePropertyShape ;
+  sh:property regen:ProjectOptionalEndDatePropertyShape ;
```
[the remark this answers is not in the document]
wgwz: Nothing special happening here except anticipating duplication (other future projects will probably have optional start/end dates). Writing `sh:property` with this type of syntax allows deduping implementation details across multiple schemas.

I.e. in the screenshot below you can see `common.ttl` schema on the left-hand side, the C01 project schema in the middle and the C03 project on the right-hand side.

Screen Shot 2022-11-01 at 12 36 44 PM

Notice how the `NamePropertyShape` is defined in one place on the left but used in multiple locations. It is "[composition](https://en.wikipedia.org/wiki/Composition_over_inheritance)" for shacl schema's.

ops/C03/project-metadata/C03-project-full-example.jsonld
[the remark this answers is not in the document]
wgwz: i think i'm going to remove all instances of `schema:Text` for the C03 schema. and at a later time we can handle switching this in other places (C01/C02):

https://github.com/regen-network/regen-registry-standards/issues/34

shacl/methodology.ttl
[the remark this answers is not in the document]
wgwz: i don't remember what @S4mmyb said about this @clevinson so i'll need input here. at least the way it is currently is compatible.

ops/C03/project-metadata/C03-project-full-example.jsonld
wgwz: for now, i've switched this to a string, and we can keep it this way if it's what will be best right now.

wgwz: when we get to the point of writing SPARQL queries, this is a choice that affects that. if it's a string in one place, but integer in another. there's ways to code around it obviously. but just wanted to mention it.

wgwz:
@clevinson @ryanchristo i know we've had some [discussion about a versioning scheme for registry-standards](https://github.com/regen-network/regen-registry-standards/issues/20). should we decide on a lightweight versioning scheme before going live with this toucan metadata?

ops/C03/project-metadata/C03-project-full-example.jsonld
```diff
@@ -0,0 +1,39 @@
...
+    "regen:projectEndDate": {
+      "@type": "xsd:date"
+    }
+  },
+  "@type": "regen:C03-Project",
+  "regen:toucanProjectTokenId": 123,
+  "schema:name": "VCS-123",
+  "regen:vcsProjectId": "VCS-123",
+  "regen:toucanURI": "http://toucan.earth/projects/VCS-123",
```
wgwz, line 23: @clevinson @ryanchristo if you have a realistic example of what the toucanURI's will be, i'd like to update what i've put in as dummy values for that.

ops/C03/project-metadata/C03-project-full-example.jsonld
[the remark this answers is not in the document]
blushi: I also agree this makes more sense to have a simpler JSON-LD structure and just use `xsd:string`

ops/C03/project-metadata/C03-project-full-example.jsonld
blushi: I also want to add that we show this `regen:vcsProjectId` on the project page UI, eg https://app.regen.network/projects/C01-002:
image not a big deal but this might result in some inconsistencies throughout project pages

shacl/projects/C03-project.ttl
```diff
@@ -0,0 +1,84 @@
...
+.
+
+regen:ToucanProjectTokenIDPropertyShape sh:path regen:toucanProjectTokenId ;
+  sh:datatype xsd:integer ;
+  sh:minCount 1 ;
+  sh:maxCount 1 ;
+  sh:minLength 1;
+.
+
+regen:C03MethodologyVersionShape a sh:NodeShape ;
```
blushi, line 60:
```suggestion
regen:C03-MethodologyVersionShape a sh:NodeShape ;
```

blushi: should we move this part to a new folder `shacl/methodologies` along with methodology.ttl that would be unchanged instead?

shacl/methodology.ttl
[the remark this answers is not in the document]
S4mmyb: I personally think `name` should be required and `identifier` should not. The reason for this is that not all methodologies will have an identifier, but all will have a name. If we're going to re-use this methodology shape for all credit classes I think we should adhere to that standard. In the toucan instance we should just make sure to pass the identifier, which is the only thing they provide, into the name field.

ops/C03/credit-batch-metadata/C03-batch-minimal-example.jsonld
```diff
@@ -0,0 +1,8 @@
+{
```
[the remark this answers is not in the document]
S4mmyb: +1, i think we should be using `ops` for anything which actually lives on chain

ops/C03/project-metadata/C03-project-full-example.jsonld
clevinson: My preference is consistency, the question then is what regex would we use to reliably convert `VCS-612` to `612`. @ryanchristo do you have thoughts given toucan's historical issues around consistent data? I think the problem was that sometimes they actually used `VCU-612` which doesn't really make sense and seemed to be a data cleaning error on their side.

clevinson: Maybe we just fail to bridge (as in bridge service reports an error) projects that aren't `^VCU-\d*$` ?

ops/C03/project-metadata/C03-project-full-example.jsonld
clevinson: We actually don't have a dummy value right now... My guess is that it will be indexed by `toucanProjectTokenId` instead of `vcsProjectId`, so something like `http://toucan.earth/projects/123` ?

ops/C03/credit-batch-metadata/C03-batch-minimal-example.jsonld
```diff
@@ -0,0 +1,8 @@
...
+  "@context": {
+    "regen": "http://regen.network/"
+  },
+  "@type": "regen:C03-Batch",
+  "regen:toucanVintageTokenId": 0,
```
ryanchristo, line 6: Should we use a non-zero value given this is an example?

ryanchristo:
```suggestion
  "regen:toucanVintageTokenId": 123,
```

ops/C03/credit-batch-metadata/C03-batch-minimal-example.jsonld
```diff
@@ -0,0 +1,8 @@
...
+  "regen:toucanURI": "http://toucan.earth/projects/VCS-123"
```
ryanchristo, line 7: Should we use something other than `projects` given this is for a batch?

ryanchristo:
```suggestion
  "regen:toucanURI": "http://toucan.earth/vintages/123"
```

ops/C03/project-metadata/C03-project-full-example.jsonld
ryanchristo: I think this is ok as a placeholder for now. :+1:

ops/C03/credit-batch-metadata/C03-batch-minimal-example.jsonld
S4mmyb: good catch @ryanchristo , i think so

ops/C03/project-metadata/C03-project-full-example.jsonld
```diff
@@ -0,0 +1,39 @@
...
+  "regen:approvedMethodologies": [
+    {
+      "schema:name": "Preservation Protocol - 40 years",
+      "schema:identifier": "PP40",
+      "schema:url": "https://pp.org/PP40",
+      "schema:version": "2022-10-26"
```
S4mmyb, line 29: not something to address here but i think version would most likely taking the form of semantic versioning, like 3.1.2

ops/C03/project-metadata/C03-project-full-example.jsonld
ryanchristo: I don't think we should fail but rather use VCS when VCU is provided (i.e. make the conversion on the bridge service)

```
180,VCS-584,VCS,AMS-I.C.,India,IN,,,,,
181,VCU-1882,VCU,VM0015,Peru,PE,,,,,
182,VCS-1882,VCU,VM0015,Peru,PE,,,,,
183,VCU-1764,VCU,AR-AM0014,Myanmar,MM,,,,,
184,VCS-1764,VCU,AR-AM0014,Myanmar,MM,,,,,
185,VCS-1715,VCU,VM0010,China,CN,,,,,
186,VCU-868,VCU,VM0007,Peru,PE,,,,,
187,VCU-92,VCU,ACM0002,India,IN,,,,,
188,VCS-92,VCU,ACM0002,India,IN,,,,,
```

https://registry.verra.org/app/projectDetail/VCS/584
https://registry.verra.org/app/projectDetail/VCS/1882

ryanchristo:
> @clevinson @ryanchristo i know we've had some https://github.com/regen-network/regen-registry-standards/issues/20. should we decide on a lightweight versioning scheme before going live with this toucan metadata?

We can probably address this separately. :+1:

ops/C03/project-metadata/C03-project-full-example.jsonld
ryanchristo: Opened https://github.com/rnd-inc/toucan-bridge-service/issues/242 for tracking

ops/C03/project-metadata/C03-project-full-example.jsonld
clevinson, line 23:
```suggestion
  "regen:toucanURI": "http://toucan.earth/projects/123",
```

shacl/methodology.ttl
```diff
@@ -15,22 +15,24 @@ regen:MethodologyVersionShape
...
     sh:minLength 1 ;
   ] ;
   sh:property [ 
     sh:path schema:identifier ;
     sh:datatype schema:Text ;
-    sh:minCount 1;
-    sh:maxCount 1;
+    sh:minCount 1 ;
```
clevinson, line 30:
```suggestion
```

ops/C03/credit-batch-metadata/C03-batch-minimal-example.jsonld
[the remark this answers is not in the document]
wgwz: makes sense, i was thinking about this too, glad it was caught in review. handled it via 6bf789d2bc4512702fcda3ea8277a4c725b73a22. it might be slightly unexpected the way i implemented it. but if you note the changes to `shacl_validate.py` in that commit, it might make sense why i chose to do it that way.

basically, `jsonld/` folder has been intended for "empty" templates as i understand it. so moving the `ops/C03/` doesn't quite make sense. at the same time, i wanted to be lazy about the `shacl_validate.py` script. it uses a wildcard to find out which of the jsonld files from the `ops/` folder we want to run validations against. furthermore, the "empty" jsonld files fail shacl validations, as expected. so it doesn't make sense to add an additional wildcard for the whole `jsonld/` folder.

for that reason, this is why i chose to do `mv ops/C03 jsonld/C03`. because then the wildcard expression adjustment's are super easy. maybe this ok, but it's probably still confusing. i think maybe we need to add a README into `jsonld/` to describe the layout? alternatively maybe we need a new folder called `ops-examples`?

this is predicated on the "empty" templates still being a thing that we need. i.e. form generation. if that's not the case, then perhaps we can fill these templates with dummy data, such that they can pass validations. and then we run validations against `ops` and `jsonld`.

if it seems to complicated i suggest we just go with what i did in the commit above for now, and open a ticket to improve it later.

shacl/methodology.ttl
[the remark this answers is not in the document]
wgwz: done! 5d488839ff46c707a873a7101ea7b0d02f5ffc58

ops/C03/project-metadata/C03-project-full-example.jsonld
wgwz: so will we use an integer or a string in our metadata? sounds like we can use an integer

shacl/credit-batches/C03-batch.ttl
```diff
@@ -0,0 +1,22 @@
+@prefix sh: <http://www.w3.org/ns/shacl#> .
+@prefix regen: <http://regen.network/> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+
+regen:C03-BatchShape a sh:NodeShape ;
+  sh:targetClass regen:C03-Batch ;
+
+  sh:property regen:ToucanVintageTokenIDPropertyShape ;
```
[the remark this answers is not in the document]
wgwz: thanks for the suggestion, i agree that's a good direction to go. the main reason i chose named shape, is because of copy and paste from other schema (;-) so not really anything important). took care of this via 27e22d2

shacl/projects/C03-project.ttl
wgwz: thanks for the suggestion @blushi , i think i implemented what you asked for here: a82b0b8 LMK

ops/C03/credit-batch-metadata/C03-batch-minimal-example.jsonld
wgwz: d2023cf

ops/C03/credit-batch-metadata/C03-batch-minimal-example.jsonld
wgwz: 17fa86d

ops/C03/project-metadata/C03-project-full-example.jsonld
wgwz: 0f73b82 switched to a semantic version as the example data

ops/C03/project-metadata/C03-project-full-example.jsonld
wgwz: a551330

wgwz:
> > @clevinson @ryanchristo i know we've had some #20. should we decide on a lightweight versioning scheme before going live with this toucan metadata?
> 
> We can probably address this separately. 👍

the reason i ask is because if we anticipate changes, and assuming there is a lot of metadata that gets created, and that later on we do want to add a version scheme for all the metadata, it could be a lot of work to go back and retroactively tag C03 metadata with the extra version field. but if we don't anticipate a lot of changes, or it's somehow easier than i realize to go back and fix this, maybe it's nbd (or maybe it's nbd for other reasons :-)

shacl/methodologies/C03-methodology.ttl
```diff
@@ -0,0 +1,30 @@
+@prefix schema: <http://schema.org/> .
+@prefix sh: <http://www.w3.org/ns/shacl#> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+@prefix regen: <http://regen.network/> .
+
+regen:C03-MethodologyVersionShape a sh:NodeShape ;
```
clevinson, line 6: What's the reason for us just not the `shacl/methodologies/methodology.ttl` schema below? I think that we'll likely be using only one schema for methodologies moving forward, and seems like we've more or less aligned on what it should contain, as well as which fields are optional (only having `name` be required).

shacl/methodologies/methodology.ttl
```diff
@@ -15,22 +15,23 @@ regen:MethodologyVersionShape
...
   sh:property [
     sh:path schema:name ;
     sh:datatype xsd:string ;
+    sh:minCount 1 ;
+    sh:maxCount 1 ;
     sh:minLength 1 ;
   ] ;
   sh:property [ 
     sh:path schema:identifier ;
     sh:datatype schema:Text ;
```
clevinson, line 29:
```suggestion
    sh:datatype xsd:string;
```

Shouldn't these be `xsd:string`?

shacl/methodologies/methodology.ttl
```diff
@@ -15,22 +15,23 @@ regen:MethodologyVersionShape
...
-    sh:minCount 1;
-    sh:maxCount 1;
+    sh:maxCount 1 ;
   ] ;
   sh:property [ 
     sh:path schema:version ;
     sh:datatype schema:Text ;
```
clevinson, line 34:
```suggestion
    sh:datatype xsd:string ;
```

shacl/projects/C03-project.ttl
```diff
@@ -0,0 +1,53 @@
...
+    sh:maxCount 1 ;
+    sh:minLength 1 ;
+  ] ;
+
+  sh:property [
+    sh:path regen:approvedMethodologies ;
+    sh:node dash:ListShape ;
+    sh:property [
+      sh:path ( [ sh:zeroOrMorePath rdf:rest ] rdf:first ) ;
+      sh:node regen:C03-MethodologyVersionShape ;
```
clevinson, line 30:
```suggestion
      sh:node regen:MethodologyVersionShape ;
```

shacl/projects/C03-project.ttl
```diff
@@ -0,0 +1,53 @@
...
+  sh:property regen:NamePropertyShape ;
+  sh:property regen:ToucanProjectTokenIDPropertyShape ;
+  sh:property [
+    sh:path regen:vcsProjectId ;
+    sh:datatype xsd:string ;
+    sh:minCount 1 ;
+    sh:maxCount 1 ;
+  ] ;
+  sh:property [
+    sh:path regen:toucanURI ;
```
clevinson, line 19: Should this be using the `regen:ToucanURIPropertyShape` defined in `shacl/credit-batches/common.ttl` ?

shacl/credit-batches/C03-batch.ttl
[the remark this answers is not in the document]
clevinson: Maybe I'm misunderstanding where we would need to be copying & pasting, but I don't think there's any other places where we would need to define a schema for `regen:toucanVintageTokenId` besides this file. And I can't find this used anywhere else at the moment. Is there another place where you imagine this field being referenced or used in shacl schemas?

shacl/projects/C03-project.ttl
```diff
@@ -0,0 +1,53 @@
...
+    sh:path regen:offsetGenerationMethod ;
+    sh:maxCount 1 ;
+    sh:minLength 1 ;
+    sh:datatype xsd:string ;
+  ] ;
+  sh:property regen:ProjectOptionalStartDatePropertyShape ;
+  sh:property regen:ProjectOptionalEndDatePropertyShape ;
+.
+
+regen:ToucanProjectTokenIDPropertyShape sh:path regen:toucanProjectTokenId ;
```
clevinson, line 48: I'm similarly confused as to when we might expect duplicate usage of this field in another schema.

clevinson (commented):
Few last questions, but I think this is really close!

ops/C03/project-metadata/C03-project-full-example.jsonld
blushi: I think we agreed to use an integer just as in existing metadata

shacl/methodologies/C03-methodology.ttl
blushi: methodology.ttl was already used for existing methodology data with more fields required (see comment at the beginning of this file) so I'd rather keep it unchanged (just like we did for credit-classes/, projects/ and /credit-batches) and define specific shapes like this one `CXX-Methodology...` as soon as those constraints are different like it is the case now

shacl/projects/C03-project.ttl
blushi: yeah if we're not expecting to use this somewhere else then there's no need for this

jsonld/C03/project-metadata/C03-project-full-example.jsonld
```diff
@@ -0,0 +1,39 @@
...
+      "@type": "xsd:date"
+    },
+    "regen:projectEndDate": {
+      "@type": "xsd:date"
+    }
+  },
+  "@type": "regen:C03-Project",
+  "regen:toucanProjectTokenId": 123,
+  "schema:name": "VCS-123",
+  "regen:vcsProjectId": "VCS-123",
```
blushi, line 22:
```suggestion
  "regen:vcsProjectId": 123,
```

jsonld/C03/project-metadata/C03-project-minimal-example.jsonld
```diff
@@ -0,0 +1,20 @@
...
+    "schema": "http://schema.org/",
+    "regen": "http://regen.network/",
+    "regen:approvedMethodologies": {
+      "@container": "@list"
+    }
+  },
+  "@type": "regen:C03-Project",
+  "regen:toucanProjectTokenId": 123,
+  "schema:name": "VCS-123",
+  "regen:vcsProjectId": "VCS-123",
```
blushi, line 12:
```suggestion
  "regen:vcsProjectId": 123,
```

jsonld/projects/C03-toucan-verified-carbon-standard-project.jsonld
```diff
@@ -0,0 +1,39 @@
...
+      "@type": "xsd:date"
+    },
+    "regen:projectEndDate": {
+      "@type": "xsd:date"
+    }
+  },
+  "@type": "regen:C03-Project",
+  "regen:toucanProjectTokenId": 0,
+  "schema:name": "",
+  "regen:vcsProjectId": "",
```
blushi, line 22:
```suggestion
  "regen:vcsProjectId": 0,
```

shacl/methodologies/methodology.ttl
blushi: we are using `schema:Text` in existing prod metadata (in credit class metadata: https://github.com/regen-network/regen-registry-standards/blob/6b832aae20f19fa18ac443b63e83208eda91a04d/ops/C01/credit-class-metadata/C01-verified-carbon-standard-credit-class.jsonld#L50 and https://github.com/regen-network/regen-registry-standards/blob/6b832aae20f19fa18ac443b63e83208eda91a04d/ops/C02/credit-class-metadata/C02-city-forest-credits-credit-class.jsonld#L48) so as long as we keep it this way there, I'd rather leave this unchanged
but if we want to update those credit classes metadata to just use xsd:string just like we decided to do for C03, then we can update that and create a separate task to handle the metadata change

jsonld/C03/credit-batch-metadata/C03-batch-minimal-example.jsonld
```diff
@@ -0,0 +1,8 @@
+{
+  "@context": {
```
victordelval, line 2: should include `"schema": "http://schema.org/"` in the @context ?

jsonld/C03/credit-batch-metadata/C03-batch-minimal-example.jsonld
```diff
@@ -0,0 +1,8 @@
...
+    "regen": "http://regen.network/"
+  },
+  "@type": "regen:C03-Batch",
```
victordelval, line 5: for @type the credit batches for C01 and C02 follow the pattern with `-CreditBatch` instead of `-Batch`, right?

shacl/methodologies/C03-methodology.ttl
wgwz: i already opened an issue for making this change throughout the repo: https://github.com/regen-network/regen-registry-standards/issues/34

shacl/methodologies/methodology.ttl
wgwz: i already opened an issue for making this change throughout the repo: https://github.com/regen-network/regen-registry-standards/issues/34

shacl/projects/C03-project.ttl
wgwz: but it’s a harmless way to write this either way

shacl/projects/C03-project.ttl
wgwz: why? defeats the purpose of having the C03-methodology. if we don’t use C03-methodology it will no be backwards compatible and we will have to update other prod metadata

shacl/projects/C03-project.ttl
wgwz: good catch! i think maybe we should pull this up into a higher level common module, will take care of that.

wgwz: b18aa1a6023234701758abc5bf830f99fa77c5fd

jsonld/C03/credit-batch-metadata/C03-batch-minimal-example.jsonld
wgwz: we don't need to include this here because there is no dependency on that vocabulary for describing C03 batch metadata. suppose we had wanted to use the `schema:name` property in this metadata, then we would need to include this context.

jsonld/C03/credit-batch-metadata/C03-batch-minimal-example.jsonld
wgwz: yes, good catch. thanks for that!

wgwz: a7ad6ae87404fe2cf1d9ffc595662dd868b61e57

ops/C03/project-metadata/C03-project-full-example.jsonld
wgwz: FYI - we made the switch to an integer for this

shacl/methodologies/methodology.ttl
blushi: Also good to add in the description there that this has to be done for the existing prod metadata which will also need some on chain credit class metadata update

ops/C03/credit-batch-metadata/C03-batch-minimal-example.jsonld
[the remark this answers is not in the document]
wgwz: i added a minimal readme to the jsonld folder: 092cc8a38ea0c4207f60db01bf6e6926b64f001d

shacl/projects/C03-project.ttl
clevinson: Cool thanks, makes sense

shacl/methodologies/methodology.ttl
clevinson: Resolving this as we'll address later.

clevinson (approved):
Lgtm! Thanks @wgwz

shacl/projects/C03-project.ttl
```diff
@@ -0,0 +1,42 @@
...
+    sh:datatype xsd:integer ;
+    sh:minCount 1 ;
+    sh:maxCount 1 ;
+    sh:minLength 1 ;
+  ] ;
+  sh:property regen:VCSProjectIdPropertyShape ;
+  sh:property regen:ToucanURIPropertyShape ;
+
+  sh:property [
+    sh:path regen:approvedMethodologies ;
```
blushi, line 22: I get this is coming from the spec but we used to have this stored at the credit class level
any reason for changing this @S4mmyb?
are we expecting C03 projects to have different approved methodologies?

S4mmyb (approved):

jsonld/C03/project-metadata/C03-project-minimal-example.jsonld
```diff
@@ -0,0 +1,20 @@
...
+    }
+  },
+  "@type": "regen:C03-Project",
+  "regen:toucanProjectTokenId": 123,
+  "schema:name": "VCS-123",
+  "regen:vcsProjectId": 123,
+  "regen:toucanURI": "http://toucan.earth/projects/123",
+  "regen:approvedMethodologies": [
+    {
+      "schema:name": "AMS-I.C.",
```
pheuberger, line 16: just want to make sure you're aware, we have projects with multiple methodologies on-chain. the field is unfortunately called `methodology` and the values are separated with a semicolon, e.g. `ACM0002; AMS-I.D.`
just so you know you'll have to parse it.

jsonld/projects/C03-toucan-verified-carbon-standard-project.jsonld
pheuberger: that's correct. did you get to amend this?

jsonld/C03/project-metadata/C03-project-full-example.jsonld
```diff
@@ -0,0 +1,39 @@
...
+      "schema:url": "https://pp.org/PP40",
+      "schema:version": "0.0.0"
+    }
+  ],
+  "regen:projectActivity": {
+    "schema:name": "Tree Preservation"
+  },
+  "regen:projectType": "Agriculture Forestry and Other Land Use",
+  "regen:offsetGenerationMethod": "Avoided Emissions",
+  "regen:projectStartDate": "2022-10-26",
```
pheuberger, line 37: just wanted to at least mention it. we don't have project activity, project type, mitigation type (what you call offsetGenerationMethod), project start and end dates on chain. a quick check of the go service confirms you're not using it anyway. what's the reason of having them if they're going to be empty?

jsonld/C03/credit-batch-metadata/C03-batch-minimal-example.jsonld
```diff
@@ -0,0 +1,8 @@
+{
+  "@context": {
+    "regen": "http://regen.network/"
+  },
+  "@type": "regen:C03-CreditBatch",
+  "regen:toucanVintageTokenId": 123,
+  "regen:toucanURI": "http://toucan.earth/vintages/123"
```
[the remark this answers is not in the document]
wgwz: hey @pheuberger i'm one of the devs at regen actively working on the standards repo here. thanks for the feedback.

regarding this field, that should be ok, we consider this to be an optional field. and this is just intended to be an example, so it's ok that the url isn't real.

jsonld/C03/project-metadata/C03-project-full-example.jsonld
```diff
@@ -0,0 +1,39 @@
...
+    },
+    "regen:projectStartDate": {
+      "@type": "xsd:date"
+    },
+    "regen:projectEndDate": {
+      "@type": "xsd:date"
+    }
+  },
+  "@type": "regen:C03-Project",
+  "regen:toucanProjectTokenId": 123,
```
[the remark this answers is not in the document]
wgwz: i think for the sake of this example metadata, as long as the type is correct for this field that's the all we are really aiming for. but sure we could change this to say `456`. do you think that would make it clearer?

jsonld/C03/project-metadata/C03-project-minimal-example.jsonld
wgwz: i guess we'll probably have to open an issue for this in the bridge service, does that sound right to you @ryanchristo ?

jsonld/C03/project-metadata/C03-project-full-example.jsonld
wgwz: these fields were added as a nice-to-have but they are optional for C03. the reason they are nice-to-have is because if they are present we can render these project details nicely in project pages. when i spoke with @clevinson about this a while ago it was planned that these would be empty by default but it could be possible to backfill these details if/when this info is or becomes available

jsonld/C03/project-metadata/C03-project-minimal-example.jsonld
ryanchristo: All set on the bridge service. :+1: 

https://github.com/rnd-inc/toucan-bridge-service/blob/main/metadata/convert.go#L76-L79
https://github.com/rnd-inc/toucan-bridge-service/blob/main/metadata/regen_project.go#L55-L68

jsonld/C03/credit-batch-metadata/C03-batch-minimal-example.jsonld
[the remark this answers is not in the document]
pheuberger: hey! nice to meet you :) 

sounds good to have this optional.

jsonld/C03/project-metadata/C03-project-minimal-example.jsonld
pheuberger: thanks for linking the code @ryanchristo  and good we had another look! verra changed the delimiter from comma to semicolon and so did we. your code still assumes comma but we already have semicolon everywhere so this wouldn’t work.

jsonld/C03/project-metadata/C03-project-full-example.jsonld
pheuberger: that makes sense 👌

jsonld/C03/project-metadata/C03-project-minimal-example.jsonld
pheuberger: you could be super defensive to support comma and semicolon just in case one or the other erroneously slips in but only semicolon would be fine too

ryanchristo: Thanks! https://github.com/rnd-inc/toucan-bridge-service/pull/315

jsonld/C03/project-metadata/C03-project-full-example.jsonld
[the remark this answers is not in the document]
aspiers: Yes `123` for example project token id and `456` for example vintage token id sounds great! :+1:

Issue #35, opened 2022-11-01
C02 project shacl schema does not specify the regen:offsetGenerationMethod


wgwz:
We missed this part of the SHACL schema prior to going live with C02.

regen-web changes/effects for this can be seen with the following data through the compaction in the jsonld-playground:

jsonld input:
```
{
  "@context": {
    "schema": "http://schema.org/",
    "regen": "http://regen.network/",
    "qudt": "http://qudt.org/schema/qudt/",
    "unit": "http://qudt.org/vocab/unit/",
    "xsd": "http://www.w3.org/2001/XMLSchema#",
    "regen:cfcProjectPage": {
      "@type": "schema:URL"
    },
    "regen:projectDesignDocument": {
      "@type": "schema:URL"
    },
    "schema:url": {
      "@type": "schema:URL"
    },
    "regen:previewPhoto": {
      "@type": "schema:URL"
    },
    "regen:galleryPhotos": {
      "@container": "@list",
      "@type": "schema:URL"
    },
    "qudt:unit": {
      "@type": "qudt:Unit"
    },
    "qudt:numericValue": {
      "@type": "xsd:double"
    },
    "regen:projectStartDate": {
      "@type": "xsd:date"
    },
    "regen:projectEndDate": {
      "@type": "xsd:date"
    }
  },
  "@type": "regen:C02-Project",
  "schema:name": "Buena Vista Heights Conservation Area",
  "regen:cfcProjectId": "9",
  "regen:cfcProjectPage": "https://www.cityforestcredits.org/carbon-credits/carbon-registry/pittsburgh-forest-carbon-offsets/",
  "regen:projectDesignDocument": "https://www.cityforestcredits.org/wp-content/uploads/2020/02/Buena-Vista-Heights-Project-Design-Document-1.pdf",
[...]

Issue #25, opened 2022-09-13
Add CLI flags to shacl_validate.py to run certain validations


wgwz:
Currently the `shacl_validate.py` script runs all validations for credit classes, credit batches and projects every time it is run. We would like to be able to run only specific validations like so:

```
./shacl_validate.py --credit-classes --credit-batches --projects
```

```
./shacl_validate.py --credit-classes --credit-batches
```

```
./shacl_validate.py --credit-batches
```

Ideally, we would just make use of python's built-in argparse library for this, since it would be nice not to have to install any python dependencies.

wgwz:
The default behavior of the script should be maintained, when we do the following, it should just run all validations:

```
./shacl_validate.py
```

S4mmyb:
I think using argparse for this in the future would be nice! I did a bit of this last year in the s2-ard repo if we want to use it as a model -> https://github.com/regen-network/regen-s2-ard/blob/384359299360115411380d67ed45698710156a68/src/ard.py#L422

PR #37, opened 2023-01-09
feat: project metadata separation

wgwz:
## Description

Closes: https://github.com/regen-network/regen-registry/issues/1420
Closes: https://github.com/regen-network/regen-registry-standards/issues/35
Closes: https://github.com/regen-network/regen-registry-standards/issues/25

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


ops/C02/project-metadata/C02-project-003.jsonld
```diff
@@ -39,7 +32,6 @@
   "regen:cfcProjectId": "9",
   "regen:cfcProjectPage": "https://www.cityforestcredits.org/carbon-credits/carbon-registry/pittsburgh-forest-carbon-offsets/",
   "regen:projectDesignDocument": "https://www.cityforestcredits.org/wp-content/uploads/2020/02/Buena-Vista-Heights-Project-Design-Document-1.pdf",
-  "schema:description": "Outside Pittsburgh, Allegheny Land Trust protected 124 acres of woodlands from rapid encroaching residential development in southeastern Allegheny County. The 40 year old maple, cherry and oak-hickory forest provides habitat for deer, turkey, and many species of birds. Hikers, birders, and mountain bikers will be able to explore the area, and possibly catch a glimpse of a majestic 200 year old oak tree.\n\nProtection of this forest also contributes to maintaining clean drinking water for Pittsburgh region’s residents. Located within the lower Youghiogheny River Watershed, the property is five miles upstream from the confluence with the Monongahela River.\n\nRevenue generated from the sale of carbon credits will be put towards acquisition costs, land stewardship, and future expansion of this and other conservation lands.",
   "regen:projectOperator": {
```
mhagel, line 35: This needs to be updated to `regen:projectDeveloper` according to my latest understanding

ops/C02/project-metadata/C02-project-004.jsonld
```diff
@@ -32,17 +25,13 @@
     },
     "regen:projectEndDate": {
       "@type": "xsd:date"
-    },
-    "regen:videoURL": {
-      "@type": "schema:URL"
     }
   },
   "@type": "regen:C02-Project",
   "schema:name": "Harvey Manning Park Expansion",
```
mhagel, line 31: FYI, the location on this metadata didn't yield a jurisdiction in our UI. Just something to be aware of. The review page in the project create flow uses a geocoding API to provide a user with a `jurisdiction` for the ledger TX. It relies on data from the `schema:location` object. I have suggested giving the user a way to manually enter it in this case.

shacl/projects/project-page.ttl
```diff
@@ -1,525 +1,86 @@
...
-        sh:path regen:projectDeveloper ;
-        sh:description "recommended to increase salability" ;
-        sh:minCount 1 ;
-        sh:maxCount 1 ;
-        sh:node regen:RoleDisplayShape ;
-        sh:group regen:ProjectPageEntityDisplayGroup ;
-      ]
+      sh:property regen:VCSProjectIdPropertyShape ;
+      sh:property regen:VCSProjectPagePropertyShape ;
+      sh:property regen:VCSMethodologyPropertyShape
```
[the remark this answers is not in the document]
wgwz: not sure why we put that there. at some point Mark and I must have had it there, just a mistake.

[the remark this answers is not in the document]
wgwz: we did have a reason for including VCS project ID and CFC project ID (your comment below). the thinking was that a reference to a project ID in the unanchored json-ld document is helpful context to have within that document. otherwise, when you are just looking at that document, you wonder, well which project is this a project page for?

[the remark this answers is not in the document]
wgwz: going to remove the project page and the methodology from the unanchored stuff, that was a mistake.

ops/C02/project-metadata/C02-project-003.jsonld
wgwz: resolved via 8e933e39dc0b8f2f6364e3e1b80ae7228459e547

shacl/projects/project-page.ttl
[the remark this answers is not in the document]
wgwz: i have moved the project page and methodology from unanchored into anchored. i have left the VCS/CFC project id's in the unanchored schema

via 80d34e3eef72776904c50ff8bbfeeb9dac2a9b79

shacl/projects/project-page.ttl
```diff
@@ -1,525 +1,86 @@
...
-        sh:path regen:landStewardPhoto ;
-        sh:datatype schema:URL ;
-        sh:name "Land steward photo" ;
-        sh:description "Upload a nice portrait of the land stewards and their families. This should be different from the other photos of land stewards you uploaded in the gallery above." ;
-        sh:minCount 1 ;
-        sh:maxCount 1 ;
-        sh:group regen:ProjectPageMediaGroup ;
-      ] ;
+      sh:property regen:CFCProjectIdPropertyShape ;
+      sh:property regen:CFCProjectPagePropertyShape
```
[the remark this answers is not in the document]
wgwz: handled via 80d34e3eef72776904c50ff8bbfeeb9dac2a9b79 (see this [comment](https://github.com/regen-network/regen-registry-standards/pull/37#discussion_r1072695647) too)

ops/C02/project-metadata/C02-project-004.jsonld
wgwz: is there something we should change the location to for this project? is it an issue with place name being an incomplete search term when using the API? in this case the place name is "Issaquah, Washington". i.e. should we change it to "Issaquah, Washington, USA". maybe i'm just misunderstanding

shacl/projects/project-page.ttl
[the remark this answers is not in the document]
blushi: I get your point but then what about just using the on chain project id in this case? this way, we don't need specific fields/validation rules for projects from different credit classes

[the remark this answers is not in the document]
wgwz: i don't have any concerns about using the on chain project id in this case. i'm only suggesting to add it, so that as a standalone JSON-LD document it is self-describing enough to be useful. perhaps this is something we don't need to worry about, i'm ok with removing it if it's not likely to provide any utility in the future.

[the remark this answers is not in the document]
blushi: Yeah I don't really see how we would use this outside of the context of it being part of the database, where the project has some id and even on_chain_id as standalone columns.
The other option could be to use the JSON-LD `@id` field and provide the project page IRI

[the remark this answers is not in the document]
wgwz: i.e. `@id` would reference https://app.regen.network/project/C02-002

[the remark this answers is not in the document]
wgwz: i refactored the C01/C02 project pages to make use of the `@id` field (via fcc5d03). i think this is a nice improvement, thanks for the suggestion @blushi :-)

wgwz:
hey @S4mmyb 👋 i just wanted to loop you in on a question that has come up while @blushi and i were reviewing these changes. the question is should we include an `@id` field in the project metadata? i'm going to try to explain why we asking this question.

when i was writing up the new project page/unanchored project metadata, it bothered me that there was no reference to a project id. without a project id, when you look at the JSON document for a project page independently, you are left wondering, well which project is this for? i think that this document should be more self-describing, so that's why i proposed an addition here.

@blushi made the suggestion that we could use the `@id` property and provide the project id in the form of a URI. this is nice because it provides a link between an unanchored project metadata and the on-chain project.

now we are wondering if we should do the same in the anchored project metadata. because the anchored metadata also has no reference to the on-chain project id. this would mean that both the project and project page metadata would have a link to the same node in our graph. here are some minimized examples:

A project page/unanchored project with an `@id` field (note the nice URI):
```
{
  "@context": {
    "schema": "http://schema.org/",
    "regen": "http://regen.network/",
    ...
  },
  "@type": "regen:Project-Page",
  "@id": "https://app.regen.network/project/C01-002",
  ...
}
```

Here is what we are thinking about adding to the project/anchored metadata:
```
{
  "@context": {
    "schema": "http://schema.org/",
    "regen": "http://regen.network/",
    ...
  },
  "@type": "regen:C01-Project",
  "@id": "https://app.regen.network/project/C01-002",
  ...
}
```

@blushi and i like the idea of adding this from the perspective of a more complete graph. but @blushi pointed out that this means that when we are creating a new on-chain project, from the perspective of the UI, we would need create the on-chain project first because we are depending on the incremented project id in the metadata now. so it's a bit awkward because we would need to sign two transactions to include the project id in the on-chain project metadata.

curious to hear your thoughts!

ops/C01/project-metadata/C01-project-id-1218-project-metadata.jsonld
```diff
@@ -4,56 +4,46 @@
     "regen": "http://regen.network/",
```
blushi, line 4: what about just using the on chain project id in the file name (instead of vcs id), similarly to other C02 projects

ops/C02/project-metadata/C02-project-001.jsonld
```diff
@@ -14,13 +14,6 @@
     "schema:url": {
```
blushi, line 14: maybe we could rename those file to "CXX-XXX-project.jsonld" (and "CXX-XXX-project-page.jsonld"), this way we see the on chain project id right away

shacl/projects/project-page.ttl.legacy
```diff
@@ -0,0 +1,526 @@
+@prefix dash: <http://datashapes.org/dash#> .
```
blushi, line 1: maybe rename to project-page.legacy.ttl?

blushi (commented):
looks good overall, just a few nits

shacl/projects/project-page.ttl.legacy
wgwz: the reason i chose the `.legacy` at the end is so that the validation script doesn't load the contents of this file into our global SHACL file. it searches for all `.ttl` files. i could change that of course but i kind of like that feature of the validation script (just load all turtle files..).

ops/C01/project-metadata/C01-project-id-1218-project-metadata.jsonld
wgwz: nice idea! implemented via 24b18e1

ops/C01/project-metadata/C01-001-project.jsonld
```diff
@@ -0,0 +1,74 @@
...
+  },
+  "@type": "regen:C01-Project",
+  "schema:name": "The Mai Ndombe REDD+ Project",
+  "regen:projectType": "Agriculture Forestry and Other Land Use",
+  "regen:projectActivity": {
+    "schema:name": "REDD",
+    "schema:url": "https://redd.unfccc.int/"
+  },
+  "regen:projectDeveloper": {
+    "@type": "regen:OrganizationDisplay",
```
blushi, line 35: I believe it would be simpler just to use this moving forward since we don't really make the difference between displayed and private party information anymore:
```suggestion
    "@type": "regen:Organization",
```

shacl/projects/common.ttl
```diff
@@ -116,6 +116,80 @@ regen:VCSProjectIdPropertyShape sh:path regen:vcsProjectId ;
...
+regen:CFCProjectPagePropertyShape sh:path regen:cfcProjectPage ;
+  sh:datatype schema:URL ;
+  sh:minCount 1 ;
+  sh:maxCount 1 ;
+  sh:group regen:ProjectPageMetadataGroup ;
+.
+
+regen:ProjectDeveloperPropertyShape sh:path regen:projectDeveloper ;
+  sh:node regen:ProjectRoleShape ;
+  sh:minCount 1 ;
```
blushi, line 148: it's optional
```suggestion
```

S4mmyb:
> hey @S4mmyb 👋 i just wanted to loop you in on a question that has come up while @blushi and i were reviewing these changes. the question is should we include an `@id` field in the project metadata? i'm going to try to explain why we asking this question.
> 
> when i was writing up the new project page/unanchored project metadata, it bothered me that there was no reference to a project id. without a project id, when you look at the JSON document for a project page independently, you are left wondering, well which project is this for? i think that this document should be more self-describing, so that's why i proposed an addition here.
> 
> @blushi made the suggestion that we could use the `@id` property and provide the project id in the form of a URI. this is nice because it provides a link between an unanchored project metadata and the on-chain project.
> 
> now we are wondering if we should do the same in the anchored project metadata. because the anchored metadata also has no reference to the on-chain project id. this would mean that both the project and project page metadata would have a link to the same node in our graph. here are some minimized examples:
> 
> A project page/unanchored project with an `@id` field (note the nice URI):
> 
> ```
> {
>   "@context": {
>     "schema": "http://schema.org/",
>     "regen": "http://regen.network/",
>     ...
>   },
>   "@type": "regen:Project-Page",
>   "@id": "https://app.regen.network/project/C01-002",
>   ...
> }
> ```
> 
> Here is what we are thinking about adding to the project/anchored metadata:
> 
> ```
> {
>   "@context": {
>     "schema": "http://schema.org/",
>     "regen": "http://regen.network/",
>     ...
>   },
>   "@type": "regen:C01-Project",
>   "@id": "https://app.regen.network/project/C01-002",
>   ...
> }
> ```
> 
> @blushi and i like the idea of adding this from the perspective of a more complete graph. but @blushi pointed out that this means that when we are creating a new on-chain project, from the perspective of the UI, we would need create the on-chain project first because we are depending on the incremented project id in the metadata now. so it's a bit awkward because we would need to sign two transactions to include the project id in the on-chain project metadata.
> 
> curious to hear your thoughts!

Hey hey, sorry for the late comment! I totally agree putting the project ids in the schemas make a lot of sense and honestly I'm surprised we didn't think of it sooner! I think the changes you have right now look great and we should move forward with them. In terms of how we operationally include the `projectID` field in the anchored and unanchored JSON-LD files I think we can really take two approaches:

1. **Add the `projectID` to the metadata prior to instantiation**. In this scenario the credit issuer would need just need to look at what the next `projectID` would be (as it auto-increments) and add it to the metadata before calling `create-class-project`. Though this doesn't seem like the ideal solution, I actually don't think it would be that big an issue because project creation is permissioned and we really won't have projects coming online frequently enough to potentially have conflicts. If we did we could always fix it with the `update-class-metadata` call.  
2. **Create the project on-chain without the ID field, then perform an `update-class-metadata call`.** 

I think either approach works and we can really choose which ever we prefer as I think we will be custodially managing the registration and metadata creation processes in the short term. In the longer run I would say we do option 2 and manage the experience creatively through our user story experience, though I'm not fully sure what that would look like. Thoughts @wgwz @clevinson @blushi

ops/C01/project-metadata/C01-001-project.jsonld
```diff
@@ -0,0 +1,74 @@
...
+    "schema:name": "Wildlife Works Carbon LLC",
+    "schema:description": "Wildlife Works is the world's leading REDD+ program development and management company.y",
+    "regen:showOnProjectPage": true
+  },
+  "regen:offsetGenerationMethod": "Emission Reduction",
```
blushi, line 40: ditto for all projects offsetGenerationMethod, based on https://github.com/regen-network/regen-registry-standards/issues/35#issuecomment-1419877284
```suggestion
  "regen:offsetGenerationMethod": ["Emission Reduction"],
```

ops/C01/project-metadata/C01-001-project.jsonld
```diff
@@ -0,0 +1,74 @@
...
+      "@type": "xsd:date"
+    },
+    "regen:projectEndDate": {
+      "@type": "xsd:date"
+    },
+    "regen:vcsProjectPage": {
+      "@type": "schema:URL"
+    }
+  },
+  "@type": "regen:C01-Project",
```
blushi, line 27: should we add the following based on https://github.com/regen-network/regen-registry-standards/pull/37#issuecomment-1419874388
```suggestion
  "@type": "regen:C01-Project",
  "@id": "https://app.regen.network/project/C01-001",
```

wgwz: added these for C01/C02 and have not done it for C03 because i only have example data for that

blushi (approved):

blushi:
> Hey hey, sorry for the late comment! I totally agree putting the project ids in the schemas make a lot of sense and honestly I'm surprised we didn't think of it sooner! I think the changes you have right now look great and we should move forward with them. In terms of how we operationally include the `projectID` field in the anchored and unanchored JSON-LD files I think we can really take two approaches:
> 
> 1. **Add the `projectID` to the metadata prior to instantiation**. In this scenario the credit issuer would need just need to look at what the next `projectID` would be (as it auto-increments) and add it to the metadata before calling `create-class-project`. Though this doesn't seem like the ideal solution, I actually don't think it would be that big an issue because project creation is permissioned and we really won't have projects coming online frequently enough to potentially have conflicts. If we did we could always fix it with the `update-class-metadata` call.
> 2. **Create the project on-chain without the ID field, then perform an `update-class-metadata call`.**
> 
> I think either approach works and we can really choose which ever we prefer as I think we will be custodially managing the registration and metadata creation processes in the short term. In the longer run I would say we do option 2 and manage the experience creatively through our user story experience, though I'm not fully sure what that would look like. Thoughts @wgwz @clevinson @blushi

To follow up on that conversation and share the recent discussions I had with @wgwz and @clevinson about that:
This might be challenging to implement for C03 projects that are created automatically from the bridge service. Quoting Cory here:

> I think we can at best say that this is optional. It doesn’t seem possible to reliably add this as part of auto generated metadata on the bridge service
> Also of note- we did discuss indexing onchain data into Apache Jena in rdf format during our offsite so I think long term we may actually be able to get this info in sparql queries without having the project ID in the project metadata dataset explicitly

PR #47, opened 2023-04-03
chore: add BCR to C03 project jsonld example

blushi:
## Description

ref: regen-network/regen-registry#1606

We should add Be Zero Carbon rating value and link to project anchored metadata, see examples in https://www.notion.so/regennetwork/8c3d639f3da24ffe8a7cd81db5adcd30?v=c2f8ac9017814e4580f6b6b1e322be17

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/jsonld/projects/C03-project.jsonld
+++ b/jsonld/projects/C03-project.jsonld
@@ -43,5 +43,9 @@
     ""
   ],
   "regen:projectStartDate": "",
-  "regen:projectEndDate": ""
+  "regen:projectEndDate": "",
+  "regen:beZeroCarbonRating": {
+    "schema:value": "",
+    "schema:url": ""
+  }
 }


jsonld/projects/C03-project.jsonld
```diff
@@ -43,5 +43,9 @@
     ""
   ],
   "regen:projectStartDate": "",
-  "regen:projectEndDate": ""
+  "regen:projectEndDate": "",
+  "regen:beZeroCarbonRating": {
```
blushi, line 47: only added this example for now, waiting for others feedback before moving forward with the SHACL schema updates

jsonld/projects/C03-project.jsonld
```diff
@@ -43,5 +43,9 @@
...
+    "schema:value": "",
```
blushi, line 48: e.g. "AA-"
I was debating whether to use https://schema.org/ratingValue instead or not... but I think be zero carbon rating might be too specific to be considered just as a `schema:ratingValue` which seems to be used for rating products or places for instance
the alternative being:
```json
"regen:beZeronCarbon": {
  "schema:ratingValue": "AA-",
  "schema:url": "https://bezerocarbon.com/..."
}
```

blushi: it would make sense if we expected other kinds of third-party ratings than BeZero Carbon ratings, do you think we will @erikalogie?
also we need to think of a structure that will allow us to display it easily this way on the project page in the future (for now, it will only appear in the "additional info" section): https://www.figma.com/file/seUiQM2SPt78KJiDL0Ea8m/Basic-Ecocredits-Ledger-View-%2F-Marketplace?node-id=3495-166691&t=AgavpAHXkCl5CZxo-0 (although this design is still not finalized...)
within `schema:name` value here, there's no reason to use camel case, we could just use the actual name in plain english: "BeZero Carbon"

wgwz: > it would make sense if we expected other kinds of third-party ratings than BeZero Carbon ratings

yup, this was my thinking as well.

> within schema:name value here, there's no reason to use camel case, we could just use the actual name in plain english: "BeZero Carbon" but I'm thinking using @type instead of schema:name might be more appropriate in this case, wdyt?

this sounds good to me as well. i don't have a strong opinion about using the `@type` parameter here at the moment. for example, that could be useful if the "beZeroRating" was going to be subject to some different conditions that other rating types. but if agree that all `regen:rating`'s will have a name, ratingValue and a url, and that's all we will need, then maybe we don't need the `@type`. but did you have another use in mind for the type?

erikalogie: > it would make sense if we expected other kinds of third-party ratings than BeZero Carbon ratings, do you think we will @erikalogie? also we need to think of a structure that will allow us to display it easily this way on the project page in the future (for now, it will only appear in the "additional info" section): https://www.figma.com/file/seUiQM2SPt78KJiDL0Ea8m/Basic-Ecocredits-Ledger-View-%2F-Marketplace?node-id=3495-166691&t=AgavpAHXkCl5CZxo-0 (although this design is still not finalized...) within `schema:name` value here, there's no reason to use camel case, we could just use the actual name in plain english: "BeZero Carbon" but I'm thinking using `@type` instead of `schema:name` might be more appropriate in this case, wdyt?

It is possible that we will have other third party ratings yes

blushi: the idea behind using @type over just `schema:name` is that if we want to attach some other data to it in the UI, like a logo or a description that could be defined from Sanity for instance and the value of `@type` would be referenced there
ie, we could use the same pattern that we use currently for SDGs or Impact where we would have the IRI referenced in Sanity
that being said, I think we could also just use both `@type` and `schema:name`:

```json
"regen:rating": {
  "@type": "regen:BeZeroCarbon",
  "schema:name": "BeZero Carbon",
  "schema:ratingValue": "AA-",
  "schema:url": "https://bezerocarbon.com/..."
}
```

blushi: thoughts @wgwz?
it'd be cool to come to a conclusion asap so I can generate the metadata IRI and update the toucan projects metadata accordingly by eod

wgwz: LGTM!

ops/C03/project-metadata/C03-001-project.jsonld
```diff
@@ -0,0 +1,71 @@
+{
+  "@type": "regen:C03-Project",
```
wgwz, line 2: just noting that for the C02, we included the `@id` parameter in the project and project-page metadata. not assuming we should be doing that here. just wanted to point it out. i don't think it's a critical issue by any means. and for these C03 projects it might be quite hard to populate that over the bridge 🤔

ops/C03/project-metadata/C03-003-project.jsonld
```diff
@@ -0,0 +1,66 @@
...
+  "regen:approvedMethodologies": {
+    "@type": "schema:ItemList",
+    "schema:itemListElement": [
+      {
+        "schema:name": "VM0015",
+        "schema:identifier": "VM0015"
+      }
+    ]
+  },
+  "schema:location": {
```
wgwz, line 26: just noting that we'll need to add this into the C03-project schema..

ops/C03/project-metadata/C03-001-project.jsonld
blushi: thanks for the reminder, yeah I think we had discussed that it's not really possible to do that over the bridge, although I could have done it here as part of those manual updates, let's try to add it next time an update is needed I guess but I think it's fine if we don't have it yet

ops/C03/project-metadata/C03-003-project.jsonld
blushi: I haven't updated the SHACL schemas yet, still TODO

shacl/projects/C03-project.ttl
```diff
@@ -23,4 +25,30 @@ regen:C03-ProjectShape a sh:NodeShape ;
   sh:property regen:ProjectTypeOptionalPropertyShape ;
   sh:property regen:ProjectOptionalStartDatePropertyShape ;
   sh:property regen:ProjectOptionalEndDatePropertyShape ;
+
+  sh:property [
```
blushi, line 29: we could extract this later when it'll be on projects from other credit classes

shacl/projects/C03-project.ttl
```diff
@@ -23,4 +25,30 @@ regen:C03-ProjectShape a sh:NodeShape ;
...
+    sh:path regen:rating ;
+    sh:node regen:BeZeroCarbon ;
```
blushi, line 31: as soon as we support more types of ratings, we could add some `sh:or` constraint here I guess instead

wgwz (approved):
tACK

PR #54, opened 2023-05-31
Schema.regen.network browser & RDFS Ontology POC

clevinson:
## Description

This PR sets up a basic ontology file in `rdfs/regen.ttl` and a corresponding next.js application that can be used to render and explore an RDFS ontology for RDF datasets and standards used within Regen Network.

To run locally:
```sh
cd schema-www
yarn install
yarn dev
```

**Warning**: Most of this code was written in collaboration with chatGPT, this is my first time coding a next.js app, or using tailwind. Please forgive me if there are antipatterns in this PR, I'll happily research and fix anything that folks point out as a problem or something sub-optimal with the way this code is currently laid out.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


rdfs/regen.ttl
```diff
@@ -0,0 +1,84 @@
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
```
clevinson, line 1: This is mostly a dummy file with some basic data to see what setting up a website that renders based on RDFS ontologies would look like.

After further research, I do think that it might be best for us to lean into OWL ontologies instead, but curious to hear others' feedback on that.

schema-www/src/styles.module.css
```diff
@@ -0,0 +1,60 @@
+/* styles.module.css */
```
clevinson, line 1: Seems like a mix of patterns to be just having a .css file with tailwind `@apply` tags everywhere...

Is there a better way for me to approach styling that's more closely adhering to tailwind best practices?

Would having these styles in-line in my component/tsx files be a preferred way to manage styles?  Since some of the styles are used throughout different components, it felt like putting it in a central file here was actually a better approach, but given the current usage I wonder if I even get anything from tailwind at all or if I should just be using vanilla css...

schema-www/src/pages/[item].tsx
```diff
@@ -0,0 +1,144 @@
...
+import ClassView from "../components/ClassView";
+import PropertyView from "../components/PropertyView";
+import { Property, Class } from "../types";
+
+// Define namespaces
+const BASE_URI = "http://schema.regen.network";
+const RDF = rdf.Namespace("http://www.w3.org/1999/02/22-rdf-syntax-ns#");
+const RDFS = rdf.Namespace("http://www.w3.org/2000/01/rdf-schema#");
+const REGEN = rdf.Namespace(BASE_URI + "#");
+const XSD = rdf.Namespace("http://www.w3.org/2001/XMLSchema#");
```
Comment (clevinson) [inline schema-www/src/pages/[item].tsx:18]: These should probably be pulled out into a separate file, and leveraged in the utils `renderLink` function as well.

schema-www/src/pages/_app.tsx
```diff
@@ -0,0 +1,16 @@
+import Link from "next/link";
+import "../styles/globals.css";
+
+function MyApp({ Component, pageProps }) {
```
flagrede, line 4: Since this project is starting fresh without any legacy constraint it would be best to use the `app router` instead of the `pages` one. It would give you better performance out of the box, more options to personalize the user experience, and a better DX (once you're familiar with React Server Component).
The doc about the `app` router is great and can be viewed here (now the default option): https://nextjs.org/docs

schema-www/src/components/ClassView.tsx
```diff
@@ -0,0 +1,44 @@
...
+            <tr>
+              <th>Property</th>
+              <th>Expected Types</th>
+              <th>Description</th>
+            </tr>
+          </thead>
+          <tbody>
+            {properties.map((prop: Property, i: number) => {
+              return (
+                <tr key={i}>
```
flagrede, line 30: It's best not to use the `index` as a key whenever possible but rather a prop related to the item (maybe `label` here). If for some reason you list order changes later on (with a sort for example), React will have to unmount and mount child elements that have changed.

schema-www/tailwind.config.js
```diff
@@ -0,0 +1,21 @@
...
+module.exports = {
+  content: [
+    "./src/pages/**/*.{js,ts,jsx,tsx,mdx}",
+    "./src/components/**/*.{js,ts,jsx,tsx,mdx}",
+    "./src/app/**/*.{js,ts,jsx,tsx,mdx}",
+  ],
+  theme: {
+    extend: {
+      height: {
+        "screen-100": "100vh",
```
flagrede, line 11: This class already exists by default in tailwind, it's `h-screen`.
If you're unsure if a class exists, you can check a cheatsheet like [this one](https://tailwindcomponents.com/cheatsheet/).

schema-www/src/utils/helpers.tsx
```diff
@@ -0,0 +1,33 @@
...
+import Link from "next/link";
+
+const BASE_URL = "http://schema.regen.network#";
+
+const prefixes = {
+  xsd: "http://www.w3.org/2001/XMLSchema#",
+  schema: "http://schema.org/",
+};
+
+export const renderLink = (resource: string) => {
```
flagrede, line 11: Should this be a component ?

schema-www/src/pages/[item].tsx
```diff
@@ -0,0 +1,144 @@
...
+    return (
+      <div className={styles.error}>
+        <h1 className={styles.title}>Error: Invalid schema type</h1>
+        <p className={styles.text}>
+          The provided schema type is not recognized. Please ensure it is either
+          &quot;class&quot; or &quot;property&quot;.
+        </p>
+      </div>
+    );
+  }
```
Comment (flagrede) [inline schema-www/src/pages/[item].tsx:141]: It's more like a convention but conditionally render item is often in JSX directly to improve readability:

```suggestion
 const isClass = item.type === "class";
 const isProperty = item.type === "property";
 const isDefault = !isClass & !isProperty
 
    return (
    <>
     {isClass && <ClassView {...item} />;}
     {isItem && <PropertyView {...item} />;}
     {isDefault && <div className={styles.error}>
        <h1 className={styles.title}>Error: Invalid schema type</h1>
        <p className={styles.text}>
          The provided schema type is not recognized. Please ensure it is either
          &quot;class&quot; or &quot;property&quot;.
        </p>
      </div>}
     </>
    );
  }
```

flagrede (commented):
Nice work! Left some comments on Tailwind, NextJS, and some React best practices.

schema-www/src/styles.module.css
haveanicedavid: Agree with @flagrede here. Best to avoid using `@apply` where possible, instead favoring react component architecture. Also if we do use `@apply`, probably best to define it in the tailwind config itself

schema-www/src/pages/_app.tsx
haveanicedavid: ^ this is one of the downside of GPT generated code - it only goes up to 2021 or whenever the training data cuts off

schema-www/src/components/ClassView.tsx
```diff
@@ -0,0 +1,44 @@
+import { Class, Property } from "../types";
+import InfoBox from "./InfoBox";
+import { renderLink, parseName } from "../utils/helpers";
+import styles from "../styles.module.css";
+
+const ClassComponent: React.FC<Class> = ({
```
haveanicedavid, line 6: IMO it's preferable to define component props within the component itself, so here defining the `Class` interface within this `ClassView` component. 

Also, `React.FC` isn't really necessary or useful anymore. It used to include a type for `children`, but that was removed in React 18...if you have `children`, you'll have to manually define them anyways (or use a helper type). Most just define react components as a normal function

schema-www/src/pages/[item].tsx
Comment (haveanicedavid) [reply schema-www/src/pages/[item].tsx:141]: I actually don't prefer the approach @flagrede outlines here just because it isn't clearly conditional what renders (IE at a glance, it's not clear if two of those conditions can be met at the same time), but it's a bit more clear if you return early on the conditional cases and avoid `else` blocks:

```tsx
  if (item.type === "class") {
    return <ClassView {...item} />;
  } 
  if (item.type === "property") {
    return <PropertyView {...item} />;
  }
  return (
    <div className={styles.error}>
      <h1 className={styles.title}>Error: Invalid schema type</h1>
      <p className={styles.text}>
        The provided schema type is not recognized. Please ensure it is either
        &quot;class&quot; or &quot;property&quot;.
      </p>
    </div>
  );
}
```

schema-www/src/types/index.ts
```diff
@@ -0,0 +1,15 @@
...
+  ranges: string[];
+  domains: string[];
+}
+export interface Class {
+  type: "class";
+  iri: string;
+  label: string;
+  description: string;
+  properties: Property[];
+}
```
haveanicedavid, line 15: Probably best to avoid a global `types` index file like this for individual component prop definitions. Works well for simple stuff, but doesn't scale

schema-www/src/pages/[item].tsx
Comment (flagrede) [reply schema-www/src/pages/[item].tsx:141]: Interesting, I didn't know we had different thoughts on this topic within the team. Regarding the point you outline @haveanicedavid if a component is meant to render only one of its children at a time then using a state that is unique instead of variables for each block clears the confusion. That way you have the same clarity while having only a single return for the component.

schema-www/src/types/index.ts
clevinson: What would you suggest instead? two type files, `Property.ts` and `Class.ts` each with one interface in a file?

schema-www/src/components/ClassView.tsx
clevinson: Is that true even if I'm using Class interface in other parts of the codebase as well? In the `[item].tsx` file I define:
```
type ItemProps = {
  item: Property | Class;
};
```

So because `Property` and `Class` interfaces are both used here and in their own component files, I had thought the interfaces should be defined elsewhere.  But I guess I could also just export the interfaces from my components (e.g. `ClassView.tsx`), and import it into `[item].tsx` from there?  That could be cleaner.  Is this what you meant?

schema-www/src/pages/_app.tsx
clevinson: Done! took a bit of digging around but I think I figured it all out.

schema-www/src/pages/[item].tsx
Comment (clevinson) [reply schema-www/src/pages/[item].tsx]: Is this what you meant @flagrede ? https://github.com/regen-network/regen-registry-standards/pull/54/commits/7a65b0365943073ad00d6d016a660eef04f2c13f

schema-www/src/app/[item]/page.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  const item = await getItem(itemSlug);
+  const componentType = item.type;
+
+  if (componentType === "class") {
+    return <ClassView {...(item as Class)} />;
+  }
+  if (item.type === "property") {
+    return <PropertyView {...(item as Property)} />;
+  }
```
Comment (flagrede) [inline schema-www/src/app/[item]/page.tsx:110]:
```suggestion
const componentType = item.type;
const isClass =  componentType === "class";
const isProperty = componentType === "property";

return <>
  {isClass && <ClassView {...(item as Class)} />}
  {isProperty && <PropertyView {...(item as Property)} />}
</>
```

schema-www/src/pages/[item].tsx
Comment (flagrede) [reply schema-www/src/pages/[item].tsx]: Not exactly, this is what I meant: https://github.com/regen-network/regen-registry-standards/pull/54/files#r1226226721
Whenever possible I try to have only one return statement by component. Here since the condition to return is based on a unique state you can be sure that you're never going to have `isClass` and `isProperty` at the same time. In case your condition is more complex (it depends on several variables) you can still use that pattern by having a util function that returns a unique state (ie: `const myState = getComponent(item.type, item.value, item.name)`. Does that make sense?

Comment (haveanicedavid) [reply schema-www/src/pages/[item].tsx]: > Whenever possible I try to have only one return statement by component
I actually prefer multiple return statements vs having conditionals live within one JSX block because I find it easier to read / understand. Not trying to argue with you here btw, I think the suggested changes you made are perfectly fine, just wanted to explain my reasoning:

If I see several `&&` statements within one JSX block, I have to check whatever those conditions are and mentally grok whether or not they could happen at the same time, conflict with each other, etc. If I see multiple returns statements, I can read them sequentially knowing that if the first condition is met, the rest won't render -- to me, it's just a bit simpler.

Like many of my code pattern opinions, it isn't really something i'm passionate about, just a preference. As long as things are easy to read (which everything you offered in feedback are!), I don't mind which approach is taken

tldr: the suggestions @flagrede is making sound good to me (even if I prefer something a bit different)

schema-www/src/components/ClassView.tsx
haveanicedavid: There's some wiggle room on things for sure, and nothing is inherently wrong with having a top-level `types.ts` file (I have some of these in the groups work, for ex), but it tends to get bloated quickly if you put too much in there so generally I'd just export the type directly from each respective component. 

These are really minor things though and easy to change later, so shouldn't prevent merging. Just offering as a general FYI

blushi (commented):
great to see this finally being worked on!

aaronc:
The biggest issue here I think is the usage of RDFS rather than reusing our existing SHACL files. Regardless of the benefits of RDFS vs OWL vs SHACL for this use case, I think we should be starting with the assumption that we're merging all the `.ttl` files in this repo and then creating a browser for them. If we think we need RDFS or OWL, we should add the relevant RDFS/OWL triples to the relevant `.ttl` files themselves in addition to the SHACL that's already there. It would be good to hear more of your thoughts on this, however, @clevinson and what your reasoning is for taking this approach.

clevinson:
This is the slack thread where we were having most discussion around this: https://regen-network.slack.com/archives/C01LX9E8QN8/p1683743649349819

As I understood it, SHACL is a constraints language designed specifically for validating JSON-LD documents, which is similar, but different, from defining vocabularies.

The main example I used in my head was that if we had a property `regen:ecosystemType` with values a finite set of potential values, there isn't really a clear way to define this in SHACL without bounding the property to a specific object (e.g. a `regen:Project` or `regen:CreditClass`).  And that we would actually need to define the constraints of ecosystem type on both the `regen:Project` and `regen:CreditClass` definitions.  

Using something like OWL or RDFS (I think actually only OWL has the ability to express finite set of values for a property), one could actually define the property `regen:ecosystemType` and then separately say which classes can have this property.

Let me know if you think I'm misunderstanding something here @aaronc

aaronc:
> This is the slack thread where we were having most discussion around this: https://regen-network.slack.com/archives/C01LX9E8QN8/p1683743649349819
> 
> As I understood it, SHACL is a constraints language designed specifically for validating JSON-LD documents, which is similar, but different, from defining vocabularies.
> 
> The main example I used in my head was that if we had a property `regen:ecosystemType` with values a finite set of potential values, there isn't really a clear way to define this in SHACL without bounding the property to a specific object (e.g. a `regen:Project` or `regen:CreditClass`). And that we would actually need to define the constraints of ecosystem type on both the `regen:Project` and `regen:CreditClass` definitions.
> 
> Using something like OWL or RDFS (I think actually only OWL has the ability to express finite set of values for a property), one could actually define the property `regen:ecosystemType` and then separately say which classes can have this property.
> 
> Let me know if you think I'm misunderstanding something here @aaronc

So I think this depends on what we actually want to accomplish.

If we want to make it possible for users to look at our documentation and create compliant credit class or project graphs, then what we want to be doing is actually documenting the SHACL constraints for those graphs.

If we just want to descriptively talk about what properties and classes exist and what their potential domains and ranges are, then documenting that with RDFS and OWL (where necessary) makes sense.

I question though how much this descriptive use case actually serves our users. It might actually be more useful to have documentation of how to write JSON-LD that will pass our SHACL constraints. That's a very different undertaking. Ideally we would have a site that does both.

clevinson:
Closing as we've moved this work over to http://github.com/regen-network/regen-data-standards

PR #56, opened 2023-06-15
feat: add support for optional credit class image

blushi:
## Description
ref: regen-network/regen-registry#1428

It also removes some unnecessary `@types`s : https://github.com/regen-network/regen-registry-standards/pull/56/commits/2872e5e755f1113615efbdcdc40890f9cec1c4b0
The impacted metadata also got updated on the prod database (their IRIs remaining the same since those changes are only format related changes and don't have any impact on the resulting canonized data)

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/jsonld/credit-classes/credit-class.jsonld
+++ b/jsonld/credit-classes/credit-class.jsonld
@@ -58,6 +58,5 @@
   "regen:verificationMethod": "",
   "regen:projectActivities": [
     ""
-  ],
-  "regen:tokenizationSource": ""
+  ]
 }
--- a/jsonld/credit-classes/credit-class.jsonld
+++ b/jsonld/credit-classes/credit-class.jsonld
@@ -22,7 +22,7 @@
       "@type": "schema:URL"
     }
   },
-  "@type": ["regen:CXX-CreditClass", "regen:CreditClass"],
+  "@type": "regen:CreditClass",
   "schema:description": "",
   "schema:name": "",
   "schema:url": "",
--- a/jsonld/credit-classes/credit-class.jsonld
+++ b/jsonld/credit-classes/credit-class.jsonld
@@ -1,0 +1,63 @@
+{
+  "@context": {
+    "schema": "http://schema.org/",
+    "regen": "https://schema.regen.network#",
+    "xsd": "http://www.w3.org/2001/XMLSchema#",
+    "schema:itemListElement": {
+      "@container": "@list"
+    },
+    "regen:sectoralScope": {
+      "@container": "@list"
+    },
+    "regen:offsetGenerationMethod": {
+      "@container": "@list"
+    },
+    "regen:projectActivities": {
+      "@container": "@list"
+    },
+    "schema:url": {
+      "@type": "schema:URL"
+    },
+    "schema:image": {
+      "@type": "schema:URL"
+    }
+  },
+  "@type": ["regen:CXX-CreditClass", "regen:CreditClass"],
+  "schema:description": "",
+  "schema:name": "",
+  "schema:url": "",
+  "schema:image": "",
+  "regen:sourceRegistry": {
+    "schema:name": "",
+    "schema:url": ""
+  },
+  "regen:sectoralScope": [
+    ""
+  ],
+  "regen:offsetGenerationMethod": [
+    ""
+  ],
+  "regen:approvedMethodologies": {
+    "@type": "schema:ItemList",
+    "schema:itemListElement": [
+      {
+        "schema:name": "",
+        "schema:url": "",
+        "schema:identifier": "",
+        "schema:version": ""
+      },
+      {
+        "schema:name": "",
+        "schema:url": "",
+        "schema:identifier": "",
+        "schema:version": ""
+      }
+    ],
+    "schema:url": ""
+  },
+  "regen:verificationMethod": "",
+  "regen:projectActivities": [
+    ""
+  ],
+  "regen:tokenizationSource": ""
+}
--- a/shacl/credit-classes/credit-class.ttl
+++ b/shacl/credit-classes/credit-class.ttl
@@ -9,7 +9,49 @@
 regen:CreditClassShape a sh:NodeShape ;
   sh:targetClass regen:CreditClass ;
   sh:property [
+    sh:path schema:name ;
+    sh:minCount 1 ;
+    sh:maxCount 1 ;
+    sh:minLength 1 ;
+    sh:datatype xsd:string ;
+  ] ;
+  sh:property [
+    sh:path schema:description ;
+    sh:minCount 1 ;
+    sh:maxCount 1 ;
+    sh:minLength 1 ;
+    sh:datatype xsd:string ;
+  ] ;
+  sh:property [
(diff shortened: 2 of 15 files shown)


jsonld/credit-classes/credit-class.jsonld
```diff
@@ -0,0 +1,63 @@
+{
```
blushi, line 1: I added this generic credit-class.jsonld template because I thought it would be useful for community credit class creation

shacl/credit-classes/credit-class.ttl
```diff
@@ -1,77 +1,15 @@
-@prefix dash: <http://datashapes.org/dash#> .
```
blushi, line 1: made the existing credit-class.ttl legacy and just documented the optional image for credit class in here
I'm wondering whether I should add properties that we're sure are generic to all credit classes too (like name, description... those needs to be confirmed first)

jsonld/credit-classes/credit-class.jsonld
```diff
@@ -0,0 +1,62 @@
...
+  "schema:url": "",
+  "schema:image": "",
+  "regen:sourceRegistry": {
+    "schema:name": "",
+    "schema:url": ""
+  },
+  "regen:sectoralScope": [
+    ""
+  ],
+  "regen:offsetGenerationMethod": [
```
blushi, line 37: probably we should start using `regen:creditGenerationMethod` instead to be more generic like discussed here: https://www.notion.so/regennetwork/Offset-Generation-Method-9d2839af4bcb42a197f019a1b131a149?pvs=4

blushi:
@wgwz could you help with the docker build issue: https://github.com/regen-network/regen-registry-standards/actions/runs/5276167983/jobs/9542523640?pr=56? thanks!

wgwz:
@blushi no prob, it was just a new release of a debian docker image that we automatically upgraded to causing the issue. i pushed a commit to pin us to the debian release that we were using previously (bullseye).

shacl/credit-classes/credit-class.ttl
wgwz: why wouldn't we want to add those @blushi ? i'm honestly just unsure either way, so thought it would be helpful to ask.

jsonld/credit-classes/credit-class.jsonld
wgwz: is it a good time to do that in this PR? or does it maybe make sense to do it all at once in a separate task? i'm just thinking of the cascade of changes we would have to make if we did do it as a part of this (and assuming we want to deprecate the `regen:offsetGenerationMethod` in general).

wgwz: but if what you had in mind was to _add_ this field here and we would still keep `regen:offsetGenerationMethod` as a field, that would be less work

shacl/credit-classes/credit-class.ttl
blushi: I was thinking that we could have one generic credit-class.ttl file that documents all the standardized fields for credit class metadata, then the non standardized fields (relative to a particular credit class) would be in separate .ttl files
Because right now, most of the constraints in C01-verified-carbon-standard-class.ttl, C02-city-forest-credits-class.ttl and C03-toucan-tco2-class.ttl are repeated

jsonld/credit-classes/credit-class.jsonld
blushi: yeah my suggestion was still to support `regen:offsetGenerationMethod` for existing credit classes so we don't have to update their metadata but use `regen:creditGenerationMethod` moving forward, especially for community credit classes. This way that doesn't imply additional work (except for some minor front-end changes).

wgwz: i'm ok with that then, makes sense to me

S4mmyb: To me credit generation method sounds more like "project activity", as in what are they doing to generate the credits. Offset generation method describes the overall method of removing or reducing emissions into the atmosphere. I think the more apt term would be `regen:carbonOffsetType` which would have the options of [ `carbon removal`, `emission reduction, `avoided emissions`[

blushi: I see, my point was to try to have something more generic that can be used for non-carbon credit classes and this is also the term we seem to be using now in the UI: https://www.figma.com/file/seUiQM2SPt78KJiDL0Ea8m/Basic-Ecocredits-Ledger-View-%2F-Marketplace?type=design&node-id=9031-253761&mode=design&t=2I1q3wMvtfKDbfMb-0 so worth discussing further, I believe we should be consistent with our field names within the metadata and the labels in the UI
cc/ @erikalogie

wgwz (approved):

shacl/credit-classes/credit-class.ttl
blushi: done

jsonld/credit-classes/credit-class.jsonld
erikalogie: Yes this naming was coming from Sarah B. but it is a bit confusing for me because on the buyers page we do use "offset generation method" which she did approve and I think we should keep it consistent. If offset generation method is more clear, then maybe we just wouldn't show that field for biodiversity credits, water credits, etc. rather than trying to make it generic? 

Btw I just noticed on the buyer's page that for ERA Brazil the project activity is listed under the "offset generation method", but I can't change it because "offset generation method" is a required field. I could just put n/a there for now, wdyt?

blushi: > Yes this naming was coming from Sarah B. but it is a bit confusing for me because on the buyers page we do use "offset generation method" which she did approve and I think we should keep it consistent. If offset generation method is more clear, then maybe we just wouldn't show that field for biodiversity credits, water credits, etc. rather than trying to make it generic?

Makes sense, in this case, should we revert to "offset generation method" on the credit class card? it's currently on dev: https://dev.app.regen.network/project/C02-001

erikalogie: Yes I guess so

S4mmyb: Non-carbon credits aren't considered offsets, so I think we should just exclude "offset generation method" for non-carbon credits. The ERABrazil example is wrong because it shouldn't be included as it's not an offset, so I think we should just put a N/A in those cases like you did @blushi. In terms of the terminology, I think "offset generation method" is fine. But if we're looking for something generic that we could put on all credit class detail cards I think I would go with Project Activity over "offset generation method" as it's something we can stick on carbon and non-carbon credit class cards.

blushi: re: ERA Brazil, @erikalogie created an issue to make the offset gen method optional for project cards on the buyers because so far they were required (from Sanity): https://github.com/regen-network/regen-registry/issues/1751

blushi: Also added https://github.com/regen-network/regen-web/issues/1952 for reverting the naming on the app

Issue #65, opened 2023-08-09
Fix C04 credit class metadata


blushi:
The C04 credit class has been created on chain with some initial metadata as specified in https://github.com/regen-network/regen-registry-standards/pull/57

But this metadata has some issues that should be fixed:
- use `regen:address` instead of `regen:walletAddress` https://github.com/regen-network/regen-registry-standards/pull/57#discussion_r1273358947
- could we just use a `@list` for `regen:bufferPoolAccounts`? if we don't expect `regen:bufferPoolAccounts` to have other properties than `schema:itemListElement` like `schema:url` or `schema:name`, then we should just have it as a list https://github.com/regen-network/regen-registry-standards/pull/57#discussion_r1273358409
- fix misusages of types/properties from schema.org as described in https://github.com/regen-network/regen-registry-standards/pull/57#discussion_r1273270388 and https://github.com/regen-network/regen-registry-standards/pull/57#discussion_r1273273058

ryanchristo:
Pending metadata for C04 from registry team.

ryanchristo:
Open questions still around co-benefits. Awaiting input from registry team.

blushi:
Missing on chain C04 update

PR #64, opened 2023-08-09
feat: Add support for storing impact/co-benefits and sdgs related info in anchored metadata

blushi:
## Description

Closes: #63 

We are still missing the actual data for the existing classes (in particular which co-benefits are measured or not) so waiting on that but the rest can be reviewed already.

---

### Author Checklist

*All items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues.*

I have...

- [x] provided a link to the relevant issue or specification
- [x] reviewed "Files changed" and left comments if necessary

### Reviewers Checklist

*All items are required. Please add a note if the item is not applicable and please **add
your handle next to the items reviewed if you only reviewed selected items**.*

I have...

- [ ] confirmed all author checklist items have been addressed
- [ ] reviewed code correctness and readability
- [ ] reviewed documentation is accurate
- [ ] manually tested (if applicable)

PR #68, opened 2023-08-29
fix: C04 metadata

blushi:
## Description

part of: #65

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### DIFF
--- a/ops/C04/credit-class-metadata/C04-ruuts-credit-class.jsonld
+++ b/ops/C04/credit-class-metadata/C04-ruuts-credit-class.jsonld
@@ -104,9 +104,9 @@
     "Integrated Cro-Livestock (ICL)"
   ],
   "regen:creditingTerm": "P10Y",
-  "regen:lookbackPeriod": "P-14Y",
+  "regen:lookbackPeriod": "-P14Y",
   "regen:permanencePeriod": "P20Y",
-  "regen:aggregatedProjects": true,
+  "regen:allowsAggregatedProjects": true,
   "regen:monitoringFrequency": "P1Y",
   "regen:verificationFrequency": "P5Y",
   "regen:verificationMethods": [
--- a/ops/C04/credit-class-metadata/C04-ruuts-credit-class.jsonld
+++ b/ops/C04/credit-class-metadata/C04-ruuts-credit-class.jsonld
@@ -39,20 +39,17 @@
     "regen:leakageApproach": {
       "@container": "@list"
     },
+    "regen:bufferPoolAccounts": {
+      "@container": "@list"
+    },
     "regen:creditingTerm": {
-      "@type": "schema:activityDuration"
+      "@type": "schema:Duration"
     },
     "regen:lookbackPeriod": {
-      "@type": "schema:activityDuration"
+      "@type": "schema:Duration"
     },
     "regen:permanencePeriod": {
-      "@type": "schema:activityDuration"
-    },
-    "regen:monitoringFrequency": {
-      "@type": "schema:frequency"
-    }, 
-    "regen:verificationFrequency": {
-      "@type": "schema:frequency"
+      "@type": "schema:Duration"
     }
   },
   "@type": "regen:C04-CreditClass",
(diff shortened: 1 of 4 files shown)


ops/C04/credit-class-metadata/C04-ruuts-credit-class.jsonld
```diff
@@ -75,8 +72,16 @@
...
+    {
+      "@id": "regen:WaterInfultration",
+      "@type": "regen:MeasuredCoBenefit",
+      "schema:name": "Water Infiltration"
+    },
+    {
+      "@id": "regen:Biodiversity",
+      "@type": "regen:MeasuredCoBenefit",
+      "schema:name": "Biodiversity"
+    }
```
blushi, line 84: still waiting on confirmation from the registry team whether these are measured or not

ops/C04/credit-class-metadata/C04-ruuts-credit-class.jsonld
```diff
@@ -75,8 +72,16 @@
...
-    "Biodiversity"
+    {
+      "@id": "regen:WaterInfultration",
+      "@type": "regen:MeasuredCoBenefit",
+      "schema:name": "Water Infiltration"
+    },
+    {
+      "@id": "regen:Biodiversity",
+      "@type": "regen:MeasuredCoBenefit",
+      "schema:name": "Biodiversity"
```
blushi, line 83: discussed this with @beccaRND and we might want to use some other term to specify it is vegetative diversity so waiting on a final decision on this https://regen-network.slack.com/archives/C05LGSWT344/p1693314623675639?thread_ts=1693313691.751109&cid=C05LGSWT344

clevinson:
@S4mmyb what do you think about us updating ` "regen:aggregatedProjects"` to `"regen:allowsAggregatedProjects"` ? I think that would be a bit more clear that this is about whether or not the credit class supports aggregated projects or not.

shacl/credit-classes/C04-ruuts-credit-class.ttl
```diff
@@ -18,17 +18,6 @@ regen:C04-CreditClassShape a sh:NodeShape ;
     sh:minCount 1 ;
     sh:maxCount 1 ;
   ] ;
-  sh:property [
-    sh:path regen:coBenefits ;
```
wgwz, line 22: as i saw this deletion i got curious about where this field is defined in our shacl.
why do we have this `regen:coBenefits` in the legacy credit class file?
https://github.com/regen-network/regen-registry-standards/blob/fix-65-C04-metadata/shacl/credit-classes/credit-class.ttl.legacy#L68-L76

i feel like i've lost track of the meaning of that term legacy means in the case.
i feel like at some point the intent was to do away with that file?
but i guess anything in that file just means that it will end up in the `credit_class.metadata` within the database.

just looking to clarify this a bit

blushi:
> @S4mmyb what do you think about us updating ` "regen:aggregatedProjects"` to `"regen:allowsAggregatedProjects"` ? I think that would be a bit more clear that this is about whether or not the credit class supports aggregated projects or not.
> 
> If we change that here, we should also change it in the fiber shed metadata.

updated: https://github.com/regen-network/regen-registry-standards/pull/68/commits/23c930ef73649a79e3c65dd1798ec60d147c716a

shacl/credit-classes/C04-ruuts-credit-class.ttl
blushi: it's defined in another PR which isn't merged yet because we're missing the actual data https://github.com/regen-network/regen-registry-standards/pull/64
but maybe we should merge it as is and have a follow-up PR for adding the metadata updates?

wgwz: @blushi ah, ok. that sgtm!

blushi: that's been merged now

wgwz:
i just noticed that even though the CI check is marked as passing, it's actually reporting some errors.
i'm not assuming that the changes here introduced those errors, and i'd guess it's some thing else recent that changed.
i haven't been paying too close attention to those because i thought they were passing!
this means something is wrong with the CI check script as well, it shouldn't be marked as passing.
maybe we need to make a separate issue for fixing that.

blushi:
> i just noticed that even though the CI check is marked as passing, it's actually reporting some errors. i'm not assuming that the changes here introduced those errors, and i'd guess it's some thing else recent that changed. i haven't been paying too close attention to those because i thought they were passing! this means something is wrong with the CI check script as well, it shouldn't be marked as passing. maybe we need to make a separate issue for fixing that.

Good catch, I've fixed those errors which were not related to validation errors but errors within the SHACL graph with missing properties (due to typos). Maybe this is why they do not make the CI fail? But in any case, I believe this should result in the CI failing indeed. I'll create a issue for that.

wgwz (approved):
thanks for fixing those issues and getting the CI checks passing again @blushi 
LGTM!

S4mmyb (approved):

